PR #23, opened 2024-04-12
adds publisher info into library

zippy:
Start the process of adding in publisher information into the app-library display.

### DIFF
--- a/src/renderer/src/applet-bundles/applet-bundles-store.ts
+++ b/src/renderer/src/applet-bundles/applet-bundles-store.ts
@@ -1,4 +1,10 @@
-import { asyncDerived, lazyLoadAndPoll, pipe, retryUntilSuccess } from '@holochain-open-dev/stores';
+import {
+  asyncDerived,
+  lazyLoad,
+  lazyLoadAndPoll,
+  pipe,
+  retryUntilSuccess,
+} from '@holochain-open-dev/stores';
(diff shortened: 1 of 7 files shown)


matthme (commented):
It seems to work fine but I'll explain how I would probably do it if I had to do it myself anyway.

I think I would put the zome call logic into the `applet-bundles-store.ts` instead of having it in the installable-applets element and instead consume it as a StoreSubscriber for the sake of reusability in other elements as well as to be able to use the `AsyncStatus` (pending/error/complete).

Two different approaches that come to mind to do this (not sure which one I would prefer, maybe rather the first one because it doesn't require polling):

1. Add a `LazyHoloHashMap` called e.g. `allPublishers` with publisher action hashes as keys in the `AppletBundlesStore` that calls the `get_publisher` zome function. Then decouple the displaying of publisher info into its own lit element that takes a publisher action hash as property and has a StoreSubscriber to above LazyHoloHashMap, i.e. `appletBundlesSture.allPublishers.get([publisher action hash])`.

2. Add a `lazyLoadAndPoll` that calls `getAllPublishers` to the `AppletBundlesStore` and then in  `installable-applets.ts` add a StoreSubscriber that subscribes to it.

matthme (approved):
Looking great! I will probably tweak the CSS a bit after merging to have the positioning be consistent.

PR #25, opened 2024-04-22
WIP: pocket as separate drawer

zippy:
Initial work to see what it feels like to have the pocket as a drawer.

### DIFF
--- a/src/renderer/src/elements/main-dashboard.ts
+++ b/src/renderer/src/elements/main-dashboard.ts
@@ -12,7 +12,7 @@ import {
 import { Hrl, mapValues } from '@holochain-open-dev/utils';
 import { notifyError, wrapPathInSvg } from '@holochain-open-dev/elements';
 import { msg } from '@lit/localize';
-import { mdiMagnify, mdiViewGalleryOutline } from '@mdi/js';
+import { mdiMagnify, mdiPocket, mdiViewGalleryOutline } from '@mdi/js';
 import {
   AppletHash,
   AppletId,
@@ -49,8 +49,10 @@ import { MossStore } from '../moss-store.js';
 import { JoinGroupDialog } from './join-group-dialog.js';
 import { CreateGroupDialog } from './create-group-dialog.js';
 
+import './search-dialog.js';
 import './pocket.js';
 import './creatable-panel.js';
+import { SearchDialog } from './search-dialog.js';
 import { MossPocket } from './pocket.js';
 import { CreatablePanel } from './creatable-panel.js';
 import { setupAppletMessageHandler } from '../applets/applet-host.js';
@@ -101,6 +103,9 @@ export class MainDashboard extends LitElement {
   @query('join-group-dialog')
   joinGroupDialog!: JoinGroupDialog;
 
+  @query('#search')
+  _searchDialog!: SearchDialog;
+
   @query('#pocket')
   _pocket!: MossPocket;
 
@@ -120,7 +125,10 @@ export class MainDashboard extends LitElement {
   _resizeDrawerX: number | null = null;
 
   @state()
-  showClipboard: boolean = false;
+  showPocket: boolean = false;
+
+  @state()
+  showSearch: boolean = false;
 
   @state()
   showCreatablePanel: boolean = false;
@@ -243,7 +251,7 @@ export class MainDashboard extends LitElement {
         this.addEventListener('cancel-select-wal', listener);
       });
     },
-    toggleClipboard: () => this.toggleClipboard(),
+    togglePocket: () => this.togglePocket(),
   };
 
   displayApplet(appletHash: AppletHash) {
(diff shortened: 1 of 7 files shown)


src/renderer/src/elements/main-dashboard.ts
```diff
@@ -1157,6 +1216,26 @@ export class MainDashboard extends LitElement {
             </button>
           </sl-tooltip>
         </div>
+        <div class="row center-content" style="margin-bottom: 5px;">
+          <sl-tooltip content="Search" placement="right" hoist>
```
matthme, line 1220: content="Pocket"

src/renderer/src/elements/main-dashboard.ts
```diff
@@ -1157,6 +1216,26 @@ export class MainDashboard extends LitElement {
...
+            <button
+              class="moss-button"
+              @click=${() => this.openPocket()}
```
matthme, line 1223: How about changing this to `togglePocket()`? I'd find it more intuitive if the pocket closes again when clicking the button twice.

src/renderer/src/elements/pocket.ts
```diff
@@ -140,14 +129,10 @@ export class MossPocket extends LitElement {
...
-        style="--width: 800px;"
-        no-header
-        @sl-initial-focus=${(e: { preventDefault: () => void }) => {
-          e.preventDefault();
-          this._searchField.focus();
-        }}
+      <sl-drawer
+        placement="bottom" class="drawer-contained"
+        id="pocket"
+        contained
```
matthme, line 135: The `contained` attribute here seems to disable close on side-click. Do you want to not have it close on side-click intentionally? If so I think I would personally prefer to have the pocket have less height in order to not cover as much tool view port.

src/renderer/src/elements/pocket.ts
```diff
@@ -157,40 +142,28 @@ export class MossPocket extends LitElement {
...
+            <img src="pocket_black.png" style="height: 30px; margin-right: 10px;">
+                ${msg('Pocket:')}
+            <div style="display:inline-block;font-size:80%;color: var(--sl-color-secondary-950)">
+              <span style="background: #e0e0e0; padding: 2px 5px; border-radius: 4px; color: black;"
+                        >Alt + P</span>
+              ${msg('to open')}
+            </div>
+          </div>
+        </div>
+        <sl-icon-button slot="header-actions"
```
matthme, line 155: A tooltip would be nice here to understand the consequences of clicking it.

src/renderer/src/elements/main-dashboard.ts
```diff
@@ -500,13 +508,25 @@ export class MainDashboard extends LitElement {
...
-            this.showClipboard = true;
+            this.showSearch = true;
+            this._searchDialog.show('open');
+            this._searchDialog.focus();
+            break;
+          case true:
+            this._searchDialog.hide();
+            break;
+        }
+      } else if (zEvent.altKey && zEvent.key === 'p') {
```
matthme, line 525: I can't press `Alt + P` with a single hand on my keyboard...I wonder whether we could choose a keyboard combination that's easy to press with a single hand.

matthme (changes requested):
Added a few comments. In general I think it's worth trying it out in a drawer and see how it feels.

we.dev.config.example.ts
```diff
@@ -86,12 +86,12 @@ export default defineConfig({
       subtitle: 'Collaborative note taking',
       description: 'Real-time notetaking based on syn',
       icon: {
-        type: 'https',
-        url: 'https://lightningrodlabs.org/projects/notebooks.png',
+        type: 'filesystem',
```
matthme, line 89: Oh, and can you undo those changes to the example config? The example config is intended to work out of the box when cloning the repository so it should not contain references to local paths. For developing on Moss I use a dedicated `we.dev.config.ts` file which is used by the `yarn applet-dev` and `yarn applet-dev-1` commands and which is in the `.gitignore` file.

src/renderer/src/elements/pocket.ts
zippy: hmmm.  I thought that you might want to be able to keep the pocket open while doing other things.  If you don't do contained that it's a modal.  But I'm not sure about that.

src/renderer/src/elements/main-dashboard.ts
zippy: Alt-e or Alt-t?

we.dev.config.example.ts
zippy: whoops that was a mistake.  Ah, that's a good idea.

src/renderer/src/elements/pocket.ts
zippy: Interesting, I couldn't get sl-tooltip to work.  I had to use `title=""`..

src/renderer/src/elements/pocket.ts
matthme: Okay that makes sense. But I have the sense that we should have it cover less vertical space then, what do you think?

PR #27, opened 2024-05-03
Debugging tools

matthme:


matthme:
@zippy merging this into `0.11.x` will break things because it contains many breaking changes already unrelated to the PR. I would rather propose to target main and then backport changes. I think backporting shouldn't be too hard.

matthme:
I'm actually completely confused now. How did commits from the 0.11.x branch get in here? I thought the initial commit from this PR was based off of the main branch. I think merging 0.11.x into this branch was a mistake.

matthme:
I'll make a new PR from commit `[6d74ba6](https://github.com/lightningrodlabs/we/pull/27/commits/6d74ba6b66080f94cd167f1ba469ce02a59683a9)` (the one before merging 0.11.x into this PR) targeting main instead.

PR #29, opened 2024-05-06
Debugging tools 2

matthme:
Moving #27 to a new PR targeting main.


matthme:
I extracted state dump and network info to dedicated components and added it to to the default apps (feedback board, tools library) as well and commited it directly to main. I will backport it next to the 0.11.x branch.

PR #62, opened 2024-08-01
make iframe stretch within constraints

LeosPrograms:
Added the following css to the embed iframe
- width: calc(100% - 10px);
- height: 100vh;

This seems to fix the embed size issues when applied from dev tools, although I did not test the changes from a fresh build.

### DIFF
--- a/libs/we-elements/src/elements/wal-embed.ts
+++ b/libs/we-elements/src/elements/wal-embed.ts
@@ -148,7 +148,7 @@ export class WalEmbed extends LitElement {
           frameborder="0"
           title="TODO"
           src="${iframeSrc}"
-          style="flex: 1; display: block; padding: 5px; margin: 0; resize: both;"
+          style="flex: 1; display: block; padding: 5px; margin: 0; resize: both; width: calc(100% - 10px); height: 100vh;"
           allow="clipboard-write;"
           @load=${() => {
             console.log('iframe loaded.');


matthme:
I think the `height: 100vh` is problematic. It makes the iframe be the size of the full viewport even if its content is much smaller. But the `width: calc(100% - 10px);` seems to solve the horizontal problem based on the limited testing I did. What's the problem that you're facing that `100vh` solves? In my very limited test scenario the height gets resized fine, it's only the width that doesn't.

LeosPrograms:
> I think the `height: 100vh` is problematic. It makes the iframe be the size of the full viewport even if its content is much smaller. But the `width: calc(100% - 10px);` seems to solve the horizontal problem based on the limited testing I did. What's the problem that you're facing that `100vh` solves? In my very limited test scenario the height gets resized fine, it's only the width that doesn't.

I was experimenting with an embedded note inside vines, and it was pretty vertically squished. Maybe wal-embed could accept a height parameter to determine the starting height?

matthme:
The vertical height is supposed to be taken into account already witth the [resizeFrameToFitContent](https://github.com/lightningrodlabs/moss/blob/feb45f707c1422b8d9f579d8e6d9249e5cd88ef4/libs/we-elements/src/elements/wal-embed.ts#L118) function that's being called once the iframe content has loaded. So I wonder whether the solution for the case of a note in Vines would actually lie in notebooks defining a min-height on its outer-most container.
It may still make sense to have the option to explicitly set the height of the wal-embed but in case of Vines I would imagine that, if possible, it would be more desirable that the height adjusts to the requirements of the iframe's content. You'd probably not want a talking stickies sticky to take up equal amounts of vertical space as a notebooks note for example.

matthme:
I think resizing has been addressed now also with https://github.com/lightningrodlabs/moss/pull/94 so I'm closing this.

PR #67, opened 2024-08-19
Feat/activity stream

LeosPrograms:
Adds a Moss-level view of activity from across all groups and tools


matthme:
Thanks a lot! I'm gonna merge this now as I'd like to use one of the functions added as part of this PR on the main branch. There is still an issue with notifications not scrolling but we can solve this separately.

PR #71, opened 2024-08-21
improve ui for notifications and activity

LeosPrograms:
Added ui changes due to feedback from @zippy

### DIFF
--- a/src/renderer/src/personal-views/activity-view/activity-asset.ts
+++ b/src/renderer/src/personal-views/activity-view/activity-asset.ts
@@ -86,7 +86,7 @@ export class ActivityAsset extends LitElement {
     return html`
       <img
         style="height: 14px; width: 14px; margin-bottom: -2px; margin-right: 3px;"
-        title="${this.renderAppletName()}"
+        title="${this.getAppletName()}"
         .src=${logo}
         alt="TODO"
       />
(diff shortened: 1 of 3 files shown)


src/renderer/src/personal-views/activity-view/activity-asset.ts
```diff
@@ -97,7 +98,7 @@ export class ActivityAsset extends LitElement {
       case 'pending':
         return html`<sl-skeleton style="height: 14px; width: 14px;" effect="pulse"></sl-skeleton> `;
       case 'complete':
-        return this.renderLogo(this.appletLogo.value.value);
+        return html`${this.renderLogo(this.appletLogo.value.value)}`;
```
matthme, line 101: I think this change is obsolete, no? `this.renderLogo()` already returns an html template string...

src/renderer/src/personal-views/activity-view/activity-asset.ts
```diff
@@ -116,7 +117,7 @@ export class ActivityAsset extends LitElement {
       case 'pending':
         return html`<div>Loading...</div>`;
       case 'complete':
-        return html`${this.appletName.value.value}`;
```
matthme, line 119: If this function is only used to get the string value, not an html template string then the `html` part should also removed from the `'pending'` and `'error'` cases I think

matthme: Right now it's a mixture between html and raw string

matthme: And if it's not used for rendering, maybe the function could just be renamed to `appletName()` instead of `renderAppletName()` but that's a detail.

LeosPrograms: removed html tag and changed to getAppletName

matthme (approved):
thanks!

PR #94, opened 2024-11-18
Fix resizing in WAL embed

zippy:
- Turns off `resize:both` to WAL iframe when not in `bare` mode so that there aren't two resize handles.
- Adds `flex` to `wal-embed` element's container so that it will resize properly.  
- Fixes collapse/un-collapse logic so that the element actually shrinks when collapsed.


matthme (approved):
Great!

PR #108, opened 2025-01-06
WIP:add foyer notification

zippy:


src/renderer/src/groups/foyer.ts
```diff
@@ -133,19 +141,23 @@ export class FoyerStore {
...
-      //         urgency: 'high',
-      //         timestamp: message.payload.created,
-      //       },
-      //     ]);
-      //   }
-      if (encodeHashToBase64(message.from) != this.myPubKeyB64) {
+      const mainWindowFocused = await window.electronAPI.isMainWindowFocused();
+      let b64From = encodeHashToBase64(message.from);
+      if (!mainWindowFocused) {
+        const profile = await toPromise(this.profilesStore.profiles.get(message.from));
```
matthme, line 147: This notification logic should probably also only happen if it's not a message from oneself, right? So only if `b64From != this.myPubKeyB64`

src/main/index.ts
```diff
@@ -949,7 +950,8 @@ if (!RUNNING_WITH_COMMAND) {
             .on('click', () => {
               console.log('Clicked on OS notification');
               createOrShowMainWindow();
-              emitToWindow(MAIN_WINDOW!, 'switch-to-applet', appletId);
+              if (appletId) emitToWindow(MAIN_WINDOW!, 'switch-to-applet', appletId);
```
matthme, line 953: What about re-using the `WeaveLocation` type here instead of taking group and appletId as separate arguments? I think it would be a bit cleaner. And I also noticed that it's actually redundant that we encode it to base64 anyway because we decode it again when handling the `switch-to-[...]` events in the frontend.

src/main/index.ts
```diff
@@ -928,6 +928,7 @@ if (!RUNNING_WITH_COMMAND) {
         notifyOS: boolean,
         appletId: AppletId | undefined,
         appletName: string | undefined,
+        group: DnaHashB64 | undefined,
```
matthme, line 931:
```suggestion
```

src/main/index.ts
matthme: We could probably also consolidate the event to `on-switch-to-weave-location` I think.

zippy: Yep good, will do.

PR #127, opened 2025-06-19
Foyer tweaks

matthme:
- add logic to only send OS notification if @mentioned in foyer notification, other wise urgency "medium"
- prettier send button


src/renderer/src/groups/foyer.ts
```diff
@@ -145,13 +145,19 @@ export class FoyerStore {
 
       if (b64From != this.myPubKeyB64) {
         if (!mainWindowFocused) {
-          const profile = await toPromise(this.profilesStore.profiles.get(message.from));
+          const senderProfile = await toPromise(this.profilesStore.profiles.get(message.from));
+          const senderNickname = senderProfile ? senderProfile.entry.nickname : b64From;
+          const myProfile = await toPromise(this.profilesStore.myProfile);
+          const myNickName = myProfile ? myProfile.entry.nickname : undefined;
```
zippy, line 151:
```suggestion
          const myNickName = myProfile ? myProfile.entry.nickname.toLowerCase() : undefined;
```

src/renderer/src/groups/foyer.ts
```diff
@@ -145,13 +145,19 @@ export class FoyerStore {
...
+
+          const amIMentioned = message.payload.text.includes(`@${myNickName}`);
```
zippy, line 153:
```suggestion
          const amIMentioned = message.payload.text.toLowerCase().includes(`@${myNickName}`);
```

zippy (changes requested):
Just need to ignore case...

PR #152, opened 2025-12-16
Feat/groupHash

zippy:
Adds groupHash to renderInfo so it can be used for example to toolInstaller()

### DIFF
--- a/src/renderer/src/applets/applet-host.ts
+++ b/src/renderer/src/applets/applet-host.ts
@@ -202,8 +202,12 @@ export function buildHeadlessWeaveClient(mossStore: MossStore): WeaveServices {
     async requestClose() {
       throw new Error('Close request is not supported in the headless WeaveClient.');
     },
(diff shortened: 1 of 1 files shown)


LeosPrograms (commented):
It looks fine to me, but I would like to defer to @ddd-mtl who added this endpoint.

src/renderer/src/applets/applet-host.ts
```diff
@@ -355,6 +355,9 @@ export async function handleAppletIframeMessage(
           (profile) => !!profile,
         ) as GroupProfile[];
 
+        // Extract the group DNA hash (there's always exactly one for applet-view)
```
ddd-mtl: I don`t believe this is true, since the IframeConfig is also storing multiple groupProfiles.

`groupStore` is the first one because its just for retrieving the authentificationToken and its possibility the same for every group.

ddd-mtl (approved):
updated the PR

PR #171, opened 2026-02-01
feat: add notification sounds with settings UI

zippy:
Add configurable notification sounds that play based on urgency level when the window is not focused. Users can configure sounds per urgency level (high/medium/low) and add custom sounds.

- Add Notifications tab to global settings with master toggle and volume control
- Include 4 built-in sounds: chime, bell, pop, ding
- Support custom sound uploads (MP3, WAV, OGG up to 500KB)
- Store settings in PersistedStore with localStorage
- Integrate sound playback in MossStore.handleNotification()
- Add translations for all 7 supported languages


src/renderer/src/services/notification-audio.ts
```diff
@@ -0,0 +1,107 @@
...
+      if (customSound) {
+        audio = new Audio(customSound.dataUrl);
+        this.audioCache.set(soundId, audio);
+      }
+    }
+
+    if (!audio) return;
+
+    try {
+      // Clone the audio element to allow overlapping plays
```
ddd-mtl: do we really want notification sounds to overlap?

src/renderer/src/elements/_new_design/moss-settings/notification-settings.ts
```diff
@@ -0,0 +1,443 @@
...
+  NotificationSoundId,
+  CustomSound,
+} from '../../../applets/types.js';
+import { notificationAudio, NotificationAudioService } from '../../../services/notification-audio.js';
+
+type UrgencyLevel = 'high' | 'medium' | 'low';
+
+@localized()
+@customElement('moss-notification-settings')
+export class MossNotificationSettings extends LitElement {
```
ddd-mtl: Would be less confusing if renamed `MossNotificationSoundSettings`, unless we expect to add more settings not related to sound in a near future.

ddd-mtl (approved):
Several duplicate hardcoded strings which could easily lead to errors if we start changing them in one place and don't follow up in the other places.
Recommend consoliding into one source of truth.
Some other recommendations in comments.

ddd-mtl (approved):

zippy:
was merged in a differnt pr

PR #176, opened 2026-02-25
Feat/hide stale agents

zippy:
adds ability for users to show and hide agents (right click, or in their profile dialog)


zippy:
sujperceeded

PR #177, opened 2026-02-26
feat: add per-user hidden agents to exclude stale peers from online-s…

zippy:
## Summary
- Adds per-user, per-group ability to hide agents from the peer list via right-click context menu or profile detail dialog
- Hidden agents are fully excluded from the networking layer: no pings sent, no pong responses, no status tracking, no online count contribution
- Hidden agents appear in a collapsible "Hidden" section at bottom of peer list with unhide option

## Test plan
- [x] Right-click an agent in the online or offline list, select "Hide agent" — agent disappears from list, online count updates
- [x] Expand "Hidden" section — hidden agent appears there
- [x] Right-click hidden agent and select "Unhide agent" — agent returns to normal list after next agentInfo poll
- [x] Click an agent to open profile detail, use Hide/Unhide button — same behavior
- [ ] Reload the app — hidden agents persist across sessions
- [x] Verify no pings are sent to hidden agents (check debug logs)


src/renderer/src/groups/elements/group-peers-status.ts
```diff
@@ -74,11 +133,16 @@ export class GroupPeersStatus extends LitElement {
     if (!this._peerStatuses.value) return html``;
     const now = Date.now();
     const myPubKey = this._groupStore.groupClient.myPubKey;
+    const myPubKeyB64 = encodeHashToBase64(myPubKey);
```
ddd-mtl, line 142: As a performance improvement, I suggest also converting all normalMembers to base64 since they are constantly used that way.

src/renderer/src/groups/group-store.ts
```diff
@@ -381,6 +393,49 @@ export class GroupStore {
...
 
+  hiddenAgents(): Readable<AgentPubKeyB64[]> {
+    return derived(this._hiddenAgents, (a) => a);
+  }
+
+  hideAgent(agent: AgentPubKey) {
+    const groupDnaHashB64 = encodeHashToBase64(this.groupDnaHash);
+    const agentB64 = encodeHashToBase64(agent);
+    let hiddenAgents = this.mossStore.persistedStore.hiddenAgents.value(groupDnaHashB64);
+    hiddenAgents.push(agentB64);
```
ddd-mtl: I'd check and return early here if agent is already hidden.

ddd-mtl (approved):
lgtm

PR #188, opened 2026-04-05
sign main-process zome calls via lair instead of cap grants

zippy:
Use a CallZomeTransform on AppWebsocket.connect that signs zome calls directly with the agent's own keypair via the lair-backed WeRustHandler, matching the pattern already used by wdocker and the applet-iframe IPC path. This removes every authorizeSigningCredentials call in the main process (index.ts and devSetup.ts), so installing a group or applet no longer writes an unnecessary CapGrant entry to the source chain, and the race between pre-authorization and concurrent renderer operations (register_and_join_applet, etc.) goes away.

Also makes weEmitter optional in signZomeCall so devSetup can call it without plumbing the emitter through, and threads weRustHandler into devSetup via launch.ts.

### DIFF
--- a/src/main/cli/devSetup.ts
+++ b/src/main/cli/devSetup.ts
@@ -18,14 +18,18 @@ import {
   AgentPubKey,
   AppWebsocket,
   AppInfo,
+  CallZomeRequest,
+  CallZomeTransform,
   DnaHashB64,
   EntryHash,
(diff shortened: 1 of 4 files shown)


LeosPrograms (approved):
This is much simpler. I tested the import/export functionality and it's still working well.

PR #189, opened 2026-04-10
Feat/unjoined tools

LeosPrograms:
- Fix and improve unjoined tools discovery and flow
- Fix group deeplink join bug
- Improve tool updating state


src/renderer/src/elements/_new_design/navigation/group-area-sidebar.ts
```diff
@@ -436,14 +429,59 @@ export class GroupAppletsSidebar extends LitElement {
                   }),
                 );
               }
-            }}
-              >
-                <div class="flex- flex-1">
-                  + ${this.numUnjoinedTools() && this.numUnjoinedTools()! > 0 ? msg('activate tools') : msg('add a tool')}
+            }}>
+              ${this.collapsed
+              ? !this.numUnjoinedTools || this.numUnjoinedTools() == 0 ? "" : html`
```
zippy: checking `!this.numUnjoinedTools` is checking whether the method exists, which it allways does. I don't think that's needed.  And maybe should be === ?

src/renderer/src/elements/_new_design/navigation/group-area-sidebar.ts
```diff
@@ -401,18 +401,11 @@ export class GroupAppletsSidebar extends LitElement {
...
-                ${msg('No tools yet.')}<br />${msg("Let's change that:")}
-              </div>
-              <button
-                class="moss-button flex flex-1"
-                style="padding-top: 10px; padding-bottom: 10px; border-radius: 10px; width: 120px; font-size: 16px;"
-                @click=${() => {
+        return this._groupApplets.value.value.size === 0 && this.groupProfileIsKnown()
+          ? html`
+          <div 
+            style="cursor: pointer;"
```
zippy: this will make the div allways be pointer even when it's not allways clickable.

src/renderer/src/elements/_new_design/moss-mini-button.ts
```diff
@@ -36,7 +36,8 @@ export class MossMiniButton extends LitElement {
     return html`<button
       style="${this.color ? `color:${this.color};border-color:${this.color};` : ''} ${this.loading
         ? 'pointer-events: none'
```
zippy, line 38: now need a ; after `pointer-events: none` so it won't make the css be `pointer-events: nonecursors...`

zippy: Or maybe this should be moved into `mossStyles` as part of the `.moss-mini-button-disabled` class instead of checking the disabled inline here?

zippy (changes requested):

src/renderer/src/elements/_new_design/navigation/group-area-sidebar.ts
LeosPrograms: This was intended as !this.numUnjoinedTools(), will change.

src/renderer/src/elements/_new_design/moss-mini-button.ts
LeosPrograms: Moving to shared styles

PR #207, opened 2026-04-30
fix: targeted reload on tool activation and enable/disable

zippy:
Fix for Issue: https://github.com/lightningrodlabs/moss/issues/191

reloadManualStores() calls groupStores.reload(), which destroys and recreates every GroupStore instance. That cascades through Lit context (group-context -> group-home), forces the groupProfile StoreSubscriber through pending, and unmounts the settings dialog and its <group-settings> element. On remount, GroupSettings defaults to its General tab — so activating or toggling a tool snapped the user back to General instead of staying on Group Tools.

Activation does not change the set of groups, and enable/disable does not change which applets are joined. Add two targeted reload helpers on MossStore that skip groupStores.reload() and only refresh the stores that actually went stale:

- reloadAfterAppletActivation: installedApps, allAppAssetInfos, and the activating group's allMyApplets / allMyInstalledApplets / allMyRunningApplets.
- reloadAfterAppletEnableDisable: installedApps, plus allMyRunningApplets for every group containing the applet (resolved via groupsForApplet).

Switch the activate flows (inactive-tools and group-home.joinNewApplet) and disableApplet/enableApplet over to these helpers. The settings dialog now stays mounted across the operation, so the tab persists.

Because the surrounding card is no longer remounted on toggle, the local appInfo in applet-settings-card and applet-detail-card no longer refreshes implicitly; refetch it after the toggle so the switch reflects the new running state.

### DIFF
--- a/src/renderer/src/elements/_new_design/group-settings/applet-settings-card.ts
+++ b/src/renderer/src/elements/_new_design/group-settings/applet-settings-card.ts
@@ -307,6 +307,13 @@ export class AppletSettingsCard extends BaseAppletSettingsCard {
           await this.mossStore.enableApplet(this.appletHash);
           notify(msg('Tool enabled.'));
         }
+        // Refresh local appInfo so the switch reflects the new state.
+        // Previously this happened implicitly because the surrounding
+        // dialog was unmounted/remounted by reloadManualStores().
+        const [appletClient] = await this.mossStore.getAppClient(
+          appIdFromAppletHash(this.appletHash),
+        );
+        this.appInfo = await appletClient.appInfo();
       }}
         >
         </sl-switch>
--- a/src/renderer/src/elements/_new_design/group-settings/applet-settings-card.ts
+++ b/src/renderer/src/elements/_new_design/group-settings/applet-settings-card.ts
@@ -308,8 +308,6 @@ export class AppletSettingsCard extends BaseAppletSettingsCard {
           notify(msg('Tool enabled.'));
         }
         // Refresh local appInfo so the switch reflects the new state.
-        // Previously this happened implicitly because the surrounding
-        // dialog was unmounted/remounted by reloadManualStores().
         const [appletClient] = await this.mossStore.getAppClient(
           appIdFromAppletHash(this.appletHash),
         );
--- a/src/renderer/src/groups/elements/applet-detail-card.ts
+++ b/src/renderer/src/groups/elements/applet-detail-card.ts
@@ -531,6 +531,13 @@ export class AppletDetailCard extends LitElement {
           await this.mossStore.enableApplet(this.appletHash);
           notify(msg('Tool enabled.'));
         }
+        // Refresh local appInfo so the switch reflects the new state.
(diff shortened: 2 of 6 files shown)


src/renderer/src/elements/_new_design/group-settings/applet-settings-card.ts
```diff
@@ -307,6 +307,13 @@ export class AppletSettingsCard extends BaseAppletSettingsCard {
           await this.mossStore.enableApplet(this.appletHash);
           notify(msg('Tool enabled.'));
         }
+        // Refresh local appInfo so the switch reflects the new state.
+        // Previously this happened implicitly because the surrounding
+        // dialog was unmounted/remounted by reloadManualStores().
```
ddd-mtl: I believe comment explaining code change shouldn't be in the code itself but in the git logs or PR comments. So here I would only keep the first comment line.

src/renderer/src/groups/elements/applet-detail-card.ts
```diff
@@ -531,6 +531,13 @@ export class AppletDetailCard extends LitElement {
           await this.mossStore.enableApplet(this.appletHash);
           notify(msg('Tool enabled.'));
         }
+        // Refresh local appInfo so the switch reflects the new state.
+        // Previously this happened implicitly because the surrounding
+        // dialog was unmounted/remounted by reloadManualStores().
```
ddd-mtl: same here

src/renderer/src/moss-store.ts
```diff
@@ -1896,6 +1896,43 @@ export class MossStore {
...
+  /**
+   * Targeted reload after activating (joining) an existing applet in a group.
+   *
+   * Unlike reloadManualStores(), this does NOT call groupStores.reload(), which
+   * would destroy and recreate every GroupStore instance and cause the
+   * surrounding UI (settings dialogs, sidebars, etc.) to unmount/remount via
+   * Lit context invalidation. Activation does not change the set of groups, so
+   * the existing GroupStore for the activating group is still valid; only its
+   * joined-applet state and the conductor's installed-app list need to refresh.
+   */
```
ddd-mtl, line 1903: I believe function comment should simply explain intent, i.e. what its trying to achieve, instead of what its not doing or what its doing differently from other functions.

src/renderer/src/moss-store.ts
```diff
@@ -1896,6 +1896,43 @@ export class MossStore {
...
+
+  /**
+   * Targeted reload after enabling/disabling an existing applet.
+   *
+   * Same rationale as reloadAfterAppletActivation: avoid groupStores.reload()
+   * so the surrounding settings UI does not unmount. Enable/disable does not
+   * change which applets are joined or which groups exist; it only changes the
+   * conductor's running-app set, so refresh installedApps and the per-group
+   * allMyRunningApplets for every group containing this applet.
+   */
```
ddd-mtl, line 1916: same here

ddd-mtl (approved):
Just some comments on the comments :)

PR #209, opened 2026-05-04
Feat/import seed forking

LeosPrograms:
- Allow passing `--seed-fork` CLI parameter on launch to add a string to each group seed on import, to prevent source chain forking when using multiple profiles on the same major Moss version.
- Skip wait for sync on import to speed process.
- Skip tool install on import when not progenitor in stewarded group.

### DIFF
--- a/src/main/cli/cli.ts
+++ b/src/main/cli/cli.ts
@@ -33,6 +33,8 @@ export const APPLET_DEV_TMP_FOLDER_PREFIX = 'moss-applet-dev';
 
 export interface CliOpts {
   profile?: string;
+  seedFork?: string;
+  seedfork?: string;
   devConfig?: string | undefined;
   devDataDir?: string | undefined;
   agentIdx?: number | undefined;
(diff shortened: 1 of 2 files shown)


src/main/index.ts
```diff
@@ -1800,6 +1805,12 @@ if (!RUNNING_WITH_COMMAND) {
     // Execute a groups import from a parsed array. Used by both dialog-based and
     // auto-import (pending) flows.
     const runGroupsImport = async (groups: GroupExportEntry[]): Promise<ImportResult[]> => {
+      const forkImportedSeed = (seed: string | undefined): string | undefined => {
+        if (!seed || !RUN_OPTIONS.seedFork) return seed;
```
ddd-mtl: `return undefined` would make it easier to read and understand the code. And probably safer if ever seed = null

src/main/cli/cli.ts
```diff
@@ -33,6 +33,8 @@ export const APPLET_DEV_TMP_FOLDER_PREFIX = 'moss-applet-dev';
 
 export interface CliOpts {
   profile?: string;
+  seedFork?: string;
+  seedfork?: string;
```
ddd-mtl: I find this quite weird. I'd recommend having only one option, or at least make sure user can't input both options.

LeosPrograms:
New approach in latest commit

PR #217, opened 2026-05-28
Feature/group dashboard

zippy:
- add a composable dashboard space for group home


zippy:
# Review — `feature/group-dashboard` (commits 0f5f4c3d → d52d5a44)

Scope: dashboard work only. Everything before `0f5f4c3d` is already merged on other branches and out of scope.

## Overview

Replaces the markdown group-home description with a tile-based Gridstack dashboard (markdown / image / iframe / WAL tiles), persisted as a `GroupMetaData` entry named `dashboard`. Adds a draggable edit-mode palette, per-tile chrome (edit/lock/color/fill-height/delete), bounded drag auto-scroll, a `validate-media-url` IPC, wal-embed `bare` mode fixes, and an e2e smoke test.

## Correctness

- **Data-URL images are validated and rejected.** [src/main/index.ts:1142](../../src/main/index.ts) rejects everything except `http(s):`. But [tests/e2e/smoke/11.group-dashboard.spec.ts:113-117](../../tests/e2e/smoke/11.group-dashboard.spec.ts#L113-L117) seeds the image tile with a `data:image/png;base64,…` URL and expects the tile to land. Either the test is currently failing, or `_validateMediaUrl` is being skipped somehow — run the smoke locally and reconcile (allow `data:` for images, or change the test).
- **Iframe sandbox is effectively no sandbox.** [group-dashboard.ts:1175](../../src/renderer/src/groups/elements/group-dashboard.ts#L1175) uses `sandbox="allow-scripts allow-same-origin allow-forms allow-popups"`. `allow-scripts allow-same-origin` together is documented as "effectively removes the sandbox" if the iframe loads same-origin content. Since the steward picks the URL, drop `allow-same-origin` unless there's a concrete reason.
- **`X-Frame-Options` is checked, but `Content-Security-Policy: frame-ancestors` is not.** Modern sites use the latter; we'll accept URLs that browsers will still refuse to frame. Either add a CSP `frame-ancestors` check in `validate-media-url`, or remove the embed validation and surface load failures inline — the current check gives false confidence.
- **SSRF in `validate-media-url`.** [src/main/index.ts:1156](../../src/main/index.ts#L1156) does a `net.fetch(rawUrl)` from the main process with `redirect: 'follow'`. Any steward can probe `http://127.0.0.1:<port>`, link-local, RFC1918 addresses, or internal services and learn response headers/status from the host's network. Reject private/loopback hosts after URL parse.
- **`net.fetch` is unbounded.** No timeout, no max-body. The body cancel is also only called on the success branch (after `if (!res.ok)` early return the socket isn't drained). Add an `AbortController` with a few-second timeout and drain on the failure path too.
- **Validation is write-time only.** Persisted URLs are loaded as-is on every render. A URL that was an image at save time can become anything later. Probably acceptable, but worth a comment so we don't mistake it for an enforcement boundary.
- **Image tiles leak viewers' IP to the image host.** Same risk for iframe URLs but it's expected there. Add `referrerpolicy="no-referrer"` on `` and `<iframe>` and/or proxy through the main process — at minimum, document it.
- **Markdown rendering relies on `markdownParseSafe`** ([src/renderer/src/utils.ts:866](../../src/renderer/src/utils.ts#L866)) — confirm that helper strips raw HTML / `<script>` / `javascript:` URLs; it's the only thing standing between a malicious steward and XSS in every member's renderer.
- **Cross-package import via `dist/`.** [group-dashboard.ts:35](../../src/renderer/src/groups/elements/group-dashboard.ts#L35) imports from `'../../../../../shared/group-client/dist/index.js'` instead of the package name. Fragile across build/link changes; use `@theweave/group-client`.
- **Polling cadence**: `groupDashboard` polls every 10s ([group-store.ts:905](../../src/renderer/src/groups/group-store.ts#L905)) — fine but make sure it's gated/paused while editing so the user's draft can't be clobbered by an inbound reload mid-edit.

## Maintainability

- **One 1811-line component** ([group-dashboard.ts](../../src/renderer/src/groups/elements/group-dashboard.ts)) mixing state, persistence, Gridstack glue, drag-autoscroll, fill-height math, add/edit dialog, palette, and rendering. Split before merge — at minimum: `dashboard-grid-controller.ts` (gridstack lifecycle + fill math + auto-scroll), `tile-dialog.ts` (the add/edit form), `dashboard-palette.ts` (the floating chip pane).
- **Seven `wip:` commits** with overlapping fixes (drag-along bug "fixed" twice, fill-height landed then reworked). Squash or rebase into coherent commits (`feat: tile-based dashboard`, `feat: dashboard edit-mode palette`, `fix: wal-embed bare sizing & cross-origin guard`, `feat: validate-media-url IPC`) — much easier to review history and bisect later.
- **Dead/no-op artifacts:** the `_saving` rationale comment at [group-dashboard.ts:110-113](../../src/renderer/src/groups/elements/group-dashboard.ts#L110-L113) describes a field that no longer exists; one of the two doc blocks above `_toggleTileFixed` ([:868-879](../../src/renderer/src/groups/elements/group-dashboard.ts#L868-L879)) is obsolete and contradicts the kept implementation. Delete both — per project rule #6 they're history, not intent.
- **`CLAUDE.md` rule #7** (measure-don't-reason) landed inside a `wip:` commit. Worth its own `docs:` commit so it isn't lost if the wip is reverted.
- **`unsafeHTML` + `markdownParseSafe`**: add an inline `// why:` pointing at the sanitizer so the next reader doesn't reach for DOMPurify on top.
- **`GroupDashboard` schema is unversioned.** Once published into a group's source chain, evolving the tile shape means a forward-compat shim. Cheap to add now — `{ schemaVersion: 1, tiles, … }` — costly to retrofit later.
- **`createRenderRoot` returns `this` (light DOM)** but the file also injects gridstack CSS via `unsafeCSS(gridstackCss)` ([:27-29](../../src/renderer/src/groups/elements/group-dashboard.ts#L27-L29)). The comment explains why for a shadow-root scenario; reconcile (either the comment is stale, or the element is actually used under group-home's shadow root, in which case the light-DOM choice is correct but the comment needs one extra sentence).

## Tests

- Smoke #11 covers markdown + image add + save + reload — good. Locks in the `Number.MAX_SAFE_INTEGER` regression explicitly.
- Missing: edit / delete / lock / fill-height / color / palette drag-drop / WAL tile / non-steward read-only path. Some (WAL, non-steward) are flagged in the spec comment as deferred to the slow suite — fine, but add a TODO commit or issue so they don't fall off.
- No test for `validate-media-url` rejection paths (bad URL, non-http, XFO deny). Easy unit-level coverage worth adding for the main-process handler.

## Suggested gates before merge

1. Reconcile the `data:` URL mismatch (test vs validator).
2. Drop `allow-same-origin` from the iframe sandbox.
3. Add SSRF guard + timeout + body-drain to `validate-media-url`.
4. Confirm `markdownParseSafe` is XSS-safe (link the sanitizer's behavior in a comment).
5. Squash the wip commits and split the 1811-line file.
6. Add a `schemaVersion` field to `GroupDashboard` before any tile entry hits a real network.

zippy:
Addressed below; pushed a cleaned-up branch (wip commits squashed into one feat: + coherent follow-ups).

Correctness

data: URL mismatch — fixed; the validator now accepts data:image/* for image tiles. While testing I also found two bigger validator bugs and fixed them: GitHub release assets are served as application/octet-stream (now accepted when the URL has an image extension), and Electron's net.fetch throws on redirect:'manual' — that was rejecting every redirecting URL. Now follows redirects.
iframe sandbox — dropped allow-same-origin; sandbox is now allow-scripts allow-forms allow-popups.
CSP frame-ancestors — added a conservative best-effort check alongside X-Frame-Options, with a comment that it's advisory.
SSRF — reject loopback/RFC1918/link-local hosts (literal IPs and DNS-resolved) before fetching. Honest limitation, documented inline: Electron's net.fetch can't re-validate redirect hops (manual throws, follow leaves res.url empty), so a public→internal redirect is still reachable, but the steward only learns ok/fail.
unbounded fetch — added a 5s AbortController timeout and body-drain on all paths.
validation write-time only — agreed, it's advisory UX, not an enforcement boundary; said so in a comment.
image IP leak — added referrerpolicy="no-referrer" to  and <iframe>; documented the residual IP exposure.
markdownParseSafe XSS — it's DOMPurify.sanitize(marked.parse(...)), which strips scripts / event handlers / javascript: URLs. Added an inline // why: pointing at the sanitizer; no other change needed.
dist/ import — now imports @theweave/group-client.
polling clobbers edits — drafts are already isolated: _activeTiles() returns _draftTiles while editing and the grid renders from drafts, so an inbound poll can't reach the edited grid (worst case a wasted re-render). No change.
Maintainability

1811-line file — split into dashboard-tile-dialog, dashboard-tile-utils, dashboard-styles, dashboard-palette (+ unit tests); group-dashboard.ts is now ~1170 lines. I deliberately did not extract the gridstack controller: it's the component's core state machine (bidirectional coupling to _draftTiles/DOM refs), so it would relocate rather than decouple, and there's no automated coverage for drag/resize/fill to catch a regression. Suggest deferring that until edit-mode tests exist.
wip commits — squashed into one feat: plus coherent review commits.
dead comments — removed (_saving, the stale _toggleTileFixed doc block).
rule #7 — now its own docs: commit (added a rule #8 too).
unsafeHTML — added the // why: sanitizer pointer.
unversioned schema — added schemaVersion: 1, stamped on write, normalized on read.
createRenderRoot/CSS comment — reconciled; clarified the shadow-root containment that motivates the scoped style block.
Tests — added validate-media-url rejection-path unit tests + tile-utils tests under a new yarn test:unit lane. The deferred edit/delete/lock/fill/WAL/non-steward e2e coverage is still open — will track separately.

PR #225, opened 2026-07-23
chore: run prettier across the repository

zippy:
Formatting-only commit produced by `yarn format` (`prettier --write .`) with the repo's prettier config. No functional changes — 320 files, whitespace/quote/markdown normalization only.

## Why

main has drifted from prettier-clean, so any contributor who runs `yarn format` (as CLAUDE.md instructs) sweeps ~300 unrelated files into their PR. This landed as part of #224 and made that PR unreviewable at 326 files. Landing the sweep here as its own commit lets #224 (and any other open branch) rebase down to its real changes.

## Verification

The formatted tree is byte-for-byte identical to the formatting portion of #224's head (diff excluding `wdocker`/`tests/src/wdocker` is empty), so #224 rebases cleanly once this merges.

## Follow-ups

- After merge, add the merged commit sha to a `.git-blame-ignore-revs` file (and `git config blame.ignoreRevsFile .git-blame-ignore-revs`) so blame skips this commit.
- Consider a CI `prettier --check .` step to prevent future drift.


zippy:
Closing unmerged — repo-wide formatting will happen on main-0.7 (the active development branch) instead, and flow to main from there. #224 will instead drop the out-of-scope files from its branch.

PR #231, opened 2026-08-05
fix(security): scope applet zome-call signing + validate WAL-window origins (Phase 0.5)

zippy:
## What this fixes

Two holes on the applet↔host trust boundary (Phase 0.5 of the maintainability plan).

**1. `sign-zome-call-applet` was unscoped.** The host signed any `(cell, zome, fn)` an applet asked for, and applets are handed the group app's auth token (for profiles access). Because a Holochain token authenticates the whole app, an installed Tool could open a websocket to the group app and drive `group` / `custom_views` / `foyer` / `assets` with the user's authority (registering tools, rewriting group metadata, reading private entries, writing asset relations). Since an applet cannot self-sign (no lair access), **scoping the signer closes the hole** without the harder per-app-token capability work.

**2. The WAL popout window had no origin validation.** It stamped every message as its own applet, and its sign path signed before any identity was established, so an embedded iframe (including a remote `https://` one) could get calls signed and misattributed.

## Approach

- `decideZomeCallSignable` (`src/main/zomeCallSigningPolicy.ts`) — pure, table-tested: sign **iff** the target is one of the applet's own cells (any zome) **or** a group cell's `profiles` zome. Everything else an applet does (assets, peer status, remote signals) already goes through the host bridge, which applies its own scoping.
- Enforced in the main process (`handleSignZomeCallApplet`), which resolves + caches the allowed cell sets from live app info. The caller's applet id(s) are passed in, **derived from the requesting iframe's origin** (a cross-group view passes every applet of its tool). Moss's own signer (`sign-zome-call`) is unchanged.
- The WAL window now derives identity via `getIframeKind`, drops messages from unrecognized origins, and uses that identity for signing and for `appletMessageToParent`.

## Verification

- `yarn typecheck` green; `yarn test:unit` green (11 new policy tests; central case: group cell + `group` zome → REJECT).
- **Adversarial review** by an independent session (bypass / spoofing / allow-list leaks / cache TOCTOU / breakage / WAL completeness): **no confirmed bypass**. Its one real finding — a 3s cache throttle that could wrongly *reject* just-installed cells — is fixed (commit 8bc0b8cc: always refresh once before refusing, coalesced).
- **Runtime e2e** (`tests/e2e/smoke/12.applet-zome-call-roundtrip.spec.ts`): real Moss + real conductor, creates a post in the example applet — signing `create_post` + `get_all_posts` to the applet's own cell — and asserts the post renders. Proven to be a real gate with a negative control (disabling the own-cell arm makes it fail; restored).

## Judgment calls — please sign off

- Enforce in **main** via origin-derived identity (the one place that signs and has admin access to resolve cells).
- Allow `profiles` on **any** installed group cell, not just the applet's own group (benign — user's own profile — and avoids an applet→group map in main).
- WAL cross-group **signing fails closed** (rejected). Believed unused; flag if a real tool hits it.
- Dev-mode `localhost` origins can self-declare identity (pre-existing behavior, dev builds only).
- The group **token handout is unchanged** by design; scoping the signer is what closes the hole.

## Scope / follow-ups

- Runtime e2e covers the **own-cell** arm. The `profiles` arm and a **cross-group** view are unit-tested but not yet exercised at runtime — can add specs if wanted.
- e2e is not in CI (only `unit-and-typecheck` + tryorama gate); #12 is a local/manual gate for now.
- `isPrivateIp` extra ranges + port restriction (assessment §4.3/§4.8) intentionally left to a separate branch (one intent per branch).

Needs a broader manual smoke test with real third-party applets before release.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -644,9 +644,12 @@ if (!RUNNING_WITH_COMMAND) {
 
     if (!appletCellCache) await rebuildAppletCellCache();
     let decision = evaluate();
-    if (!decision.sign) {
-      // Refresh once before refusing: the cell may belong to an applet or group
-      // installed since the cache was last built.
+    // Refresh once before refusing only when the cell is simply unknown — it may
+    // belong to an applet or group installed since the cache was last built. A
+    // known group cell refused for a non-profiles zome cannot become allowed by
+    // a rebuild, so don't pay a listApps round-trip for those (a misbehaving
+    // applet could otherwise trigger one per rejected call).
+    if (!decision.sign && decision.reason === 'unknown-cell') {
       await rebuildAppletCellCache();
       decision = evaluate();
     }
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -82,6 +82,8 @@ import {
   AppWebsocket,
   CallZomeRequest,
   CallZomeTransform,
+  CellId,
+  CellInfo,
   CellType,
   DnaHashB64,
   InstalledAppId,
@@ -109,11 +111,16 @@ import { type WeRustHandler } from '@lightningrodlabs/we-rust-utils';
 import {
   appletIdFromAppId,
   deriveToolCompatibilityId,
+  getCellId,
   globalPubKeyFromListAppsResponse,
   toLowerCaseB64,
   toolCompatibilityIdFromDistInfo,
   toOriginalCaseB64,
 } from '@theweave/utils';
+import {
+  decideZomeCallSignable,
+  type ZomeCallSigningDecision,
+} from './zomeCallSigningPolicy';
 import { sortVersionsDescending } from './utils';
 import { Profile as AgentProfile } from '@holochain-open-dev/profiles';
 import { Jimp } from 'jimp';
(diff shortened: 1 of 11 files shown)


zippy:
Thanks — addressed 1–4 in 868cbf49; 5–6 deferred as you framed them.

**1 (blocker) — done.** Identity derivation and the skip check moved to the top of the message listener (computed once); `get-iframe-config`/`unregister-iframe` now use that shared `iframeKind`/`derivedSource`, and their unreachable guards are gone (~40 lines removed). The source assembly is extracted to a pure `deriveWalMessageSource` (`wal-message-source.ts`) with table tests — which also gives the `groupHash ?? fallback` logic the unit coverage you flagged as missing.

**2 + 3 — fixed, one consistent way (slight pushback on the specific remedy).** Rather than throw for the not-ready case (2) and separately reword the unknown-origin throw (3), both now **skip without responding**, matching the main window's `if (!receivedFromSource) return`. That removes the spurious `{ success, undefined }` you identified in (2) *and* the misleading "Ignoring…" throw in (3), with one pattern instead of a new throw path. My reasoning for skip over fail-loud: it's exactly what the main window does for the same "not our message / not ready" conditions, so the two windows now behave identically; the not-ready window is a sub-second race where the child reissues `get-iframe-config` rather than wanting a hard error. Happy to switch to throw if you'd rather fail loud — it's a one-liner either way.

**4 — done.** `authorizeAppletZomeCall` refreshes the cache before refusing only when `reason === 'unknown-cell'`. A known group cell refused for a non-profiles zome can't be flipped by a rebuild, so a misbehaving applet's rejected-call loop no longer triggers a `listApps` each time.

**5 (uninstall cache clear) — deferred.** Agreed it's theoretical (no live iframe for an uninstalled app), and with (4) the rebuild only fires on `unknown-cell`, so a stale entry is never *reached* by a caller. Will do the one-line clear as a follow-up.

**6 (extract cache/authorizer from index.ts) — deferred.** Agreed it's the natural next extraction; keeping it off this security PR to avoid re-review churn, and it wants a fake admin-websocket to test properly. Follow-up branch.

Verified after the changes: typecheck + unit (55) green; the e2e signing round-trip (#12) still passes, and still fails under its negative control.

zippy:
Good catch — fixed in e822f4bd. Wrapped the `getIframeKind` call so its throw path (unrecognized origin, e.g. an embedded remote iframe) is caught: it now `console.warn`s and returns (fails closed) instead of escaping the async listener as an unhandled rejection. Restores the logging parity with the main window; security behavior unchanged. typecheck + unit green.

PR #233, opened 2026-08-10
Phase 1: wire-contract drift guards + F4 crash detection (+ small follow-ups)

zippy:
Phase 1 of the maintainability plan (close the wire-contract drift; add the conductor crash-detection fix), plus the small follow-ups that had accrued. One concern per commit.

## Phase 1

- **Remove 5 dead IPC channels + drift test** (`9075b5c3`). `enable-dev-mode`, `disable-dev-mode`, `fetch-icon`, `open-app`, `open-appstore` were invoked in preload with no `ipcMain.handle` (rejecting at runtime); none had a caller. `src/main/ipc-contract-drift.test.ts` now fails the build if any preload channel lacks a handler or any handler is unreachable — the realized-bug class can't recur.
- **Exhaustive applet-host switch + TypeBox↔union link** (`2a4f1363`). The `handleAppletIframeMessage` default is now a `never` exhaustiveness guard (a new `AppletToParentRequest` variant without a case is a compile error; this surfaced two legacy variants `ready`/`search` that no applet sends). `validationSchemas.ts` links to the union so the validator's message-type discriminants must match `AppletToParentRequest` (`WireContractInSync`) — verified non-vacuous with a negative control. Two of the three hand-maintained copies of the applet contract now fail the build on drift.
- **Conductor/lair crash detection, F4** (`ae86546a`). The launch promise only settled on `'Conductor ready.'` or two literal error strings and nothing watched the process — a port race or incompatible DB left the UI hung at "starting Holochain..." forever. Added `error`/`exit` listeners + a generous (5 min) ready-timeout backstop, guarded by a single `settled` flag; a post-startup crash is surfaced via `emitHolochainFatalPanic` instead of dropped.

## Follow-ups

- **wdocker network defaults + signal cleanup** (`68459818`) — the second step from #232: point wdocker's bootstrap/relay defaults at `bootstrap-relay.moss.social`, remove the unused `PRODUCTION_SIGNALING_URLS`.
- **Extract `AppletZomeCallAuthorizer` + invalidate on uninstall** (`d08cde5d`) — deferred #231 review findings 5 & 6: move ~95 lines of signing cache/authorization out of the 3,200-line `index.ts` into a unit-tested module, and clear its cache on uninstall so it never keeps granting for a gone app.

## Verification

- `yarn typecheck` + `yarn test:unit` (138) green.
- e2e `#12` passes — launches a real conductor (F4 happy path) and round-trips an applet zome call through the extracted signer.

## Not included (noted follow-ups)
- The full 78-handler typed IPC contract (the drift *test* covers the realized risk at much lower cost; a full contract-typed registration is a larger refactor for later).
- F4's non-re-entrant launch handler (orphans a conductor on double-launch) and quit ordering (lair before conductor).

### DIFF
--- a/src/main/appletZomeCallAuthorizer.ts
+++ b/src/main/appletZomeCallAuthorizer.ts
@@ -4,22 +4,26 @@ import type { AppletId } from '@theweave/api';
 import { decideZomeCallSignable, type ZomeCallSigningDecision } from './zomeCallSigningPolicy';
 
 /**
- * How long a resolved cell set may be trusted before it is rebuilt from live app
- * info. This is the self-heal window for an uninstall that never called
- * invalidate() — notably the renderer's leaveGroup, which uninstalls apps
- * through its own admin websocket and so bypasses the main-process invalidate
- * sites. Within this window a torn-down applet's cells can still be granted, but
- * there is no iframe left to request signing for them.
+ * How long a resolved cell set may be served before it is refreshed. Past this
+ * age the cache is refreshed in the background while the current (stale) set is
+ * still used to decide, so signing never blocks on — or fails with — a listApps
+ * round-trip. It is the self-heal window for an uninstall that never called
+ * invalidate(): notably the renderer's leaveGroup, which uninstalls apps through
+ * its own admin websocket and so bypasses the main-process invalidate sites. A
+ * torn-down applet's cells can still be granted for up to this window, but there
+ * is no iframe left to request signing for them.
  */
 const CACHE_MAX_AGE_MS = 60_000;
 
 /**
  * Minimum time between rebuilds triggered by an unknown-cell refusal. A rebuild
- * picks up an applet or group installed since the last build (so a legitimate
- * just-installed cell is not falsely refused), but only once per interval, so a
- * flood of calls to a genuinely-unknown cell cannot drive an unthrottled
- * listApps per call. The interval is well below the time it takes a
- * newly-installed applet's iframe to mount and make its first call.
+ * picks up an applet or group installed since the last build, but only once per
+ * interval, so a flood of calls to a genuinely-unknown cell cannot drive an
+ * unthrottled listApps per call. Legitimate new cells do not depend on this
+ * timing: the install IPC handlers call invalidate(), which drops the cache so
+ * the next call rebuilds regardless of the interval. The residual case is a cell
+ * that appears with no main-process signal (an applet cloning its own DNA over
+ * the app websocket) and is called within the interval of the last build.
  */
 const MIN_UNKNOWN_REBUILD_INTERVAL_MS = 1_000;
 
@@ -29,17 +33,18 @@ const MIN_UNKNOWN_REBUILD_INTERVAL_MS = 1_000;
  * decideZomeCallSignable policy).
  *
  * Resolving the allowed cells is a listApps round-trip and signing is on the hot
- * path, so the result is cached. Freshness is maintained three ways: invalidate()
- * drops it immediately (called on the IPC uninstall paths); a request that the
- * cached view refuses for an *unknown* cell rebuilds once (throttled) so a
- * just-installed cell is picked up; and a cache older than CACHE_MAX_AGE_MS is
- * rebuilt before use so a missed invalidate self-heals. A generation counter
- * ensures an invalidate() during an in-flight rebuild discards that rebuild's
- * (now stale) snapshot instead of silently restoring it.
+ * path, so the result is cached. Freshness is maintained four ways: invalidate()
+ * drops it immediately (called on the app install/uninstall IPC paths); a request
+ * refused for an *unknown* cell rebuilds once (throttled) so a just-installed
+ * cell is picked up; a cache older than CACHE_MAX_AGE_MS is refreshed in the
+ * background so a missed invalidate self-heals without blocking signing; and a
+ * generation counter makes an invalidate() during an in-flight rebuild discard
+ * that rebuild's (now stale) snapshot instead of committing it.
  */
 export class AppletZomeCallAuthorizer {
-  private cache: { ownByAppletId: Map<string, Set<string>>; groupCells: Set<string> } | undefined;
-  private builtAt = 0;
+  private cache:
+    | { ownByAppletId: Map<string, Set<string>>; groupCells: Set<string>; builtAt: number }
+    | undefined;
   private generation = 0;
   private rebuildInFlight: Promise<void> | undefined;
 
@@ -68,12 +73,22 @@ export class AppletZomeCallAuthorizer {
         groupCellIdsB64: this.cache?.groupCells ?? new Set(),
       });
 
-    if (!this.cache || this.clock() - this.builtAt > CACHE_MAX_AGE_MS) await this.rebuild();
+    if (!this.cache) {
+      // Cold start (or just invalidated): must block until we know the cells.
+      await this.rebuild();
+    } else if (this.clock() - this.cache.builtAt > CACHE_MAX_AGE_MS) {
+      // Stale-while-revalidate: keep serving the current set and refresh in the
+      // background. The .catch keeps a rejected listApps (socket reconnecting,
+      // conductor restarting) from becoming an unhandled rejection or a signing
+      // outage.
+      void this.rebuild().catch(() => {});
+    }
+
     let decision = evaluate();
     if (
       !decision.sign &&
       decision.reason === 'unknown-cell' &&
-      this.clock() - this.builtAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS
+      (!this.cache || this.clock() - this.cache.builtAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS)
     ) {
       await this.rebuild();
       decision = evaluate();
@@ -98,29 +113,29 @@ export class AppletZomeCallAuthorizer {
     if (this.rebuildInFlight) return this.rebuildInFlight;
     const generationAtStart = this.generation;
     const p = (async () => {
-      const apps = await this.listApps();
-      // An invalidate() during the listApps round-trip means this snapshot may
-      // already be stale (e.g. the app it would grant was just uninstalled), so
-      // discard it rather than restore it.
-      if (generationAtStart !== this.generation) return;
-      const ownByAppletId = new Map<string, Set<string>>();
-      const groupCells = new Set<string>();
-      for (const app of apps) {
-        const appId = app.installed_app_id;
-        if (appId.startsWith('applet#')) {
-          ownByAppletId.set(appletIdFromAppId(appId), new Set(cellIdsOfApp(app)));
-        } else if (appId.startsWith('group#')) {
-          // Only the `group` role cell hosts the profiles zome an applet may call.
-          for (const key of cellIdsOfApp(app, 'group')) groupCells.add(key);
+      try {
+        const apps = await this.listApps();
+        // An invalidate() during the listApps round-trip means this snapshot may
+        // already be stale (e.g. the app it would grant was just uninstalled), so
+        // discard it rather than commit it.
+        if (generationAtStart !== this.generation) return;
+        const ownByAppletId = new Map<string, Set<string>>();
+        const groupCells = new Set<string>();
+        for (const app of apps) {
+          const appId = app.installed_app_id;
+          if (appId.startsWith('applet#')) {
+            ownByAppletId.set(appletIdFromAppId(appId), new Set(cellIdsOfApp(app)));
+          } else if (appId.startsWith('group#')) {
+            // Only the `group` role cell hosts the profiles zome an applet may call.
+            for (const key of cellIdsOfApp(app, 'group')) groupCells.add(key);
+          }
         }
+        this.cache = { ownByAppletId, groupCells, builtAt: this.clock() };
+      } finally {
+        if (generationAtStart === this.generation) this.rebuildInFlight = undefined;
       }
-      this.cache = { ownByAppletId, groupCells };
-      this.builtAt = this.clock();
     })();
     this.rebuildInFlight = p;
-    void p.finally(() => {
-      if (this.rebuildInFlight === p) this.rebuildInFlight = undefined;
-    });
     return p;
   }
 }
--- a/src/main/appletZomeCallAuthorizer.ts
+++ b/src/main/appletZomeCallAuthorizer.ts
@@ -16,14 +16,15 @@ import { decideZomeCallSignable, type ZomeCallSigningDecision } from './zomeCall
 const CACHE_MAX_AGE_MS = 60_000;
 
 /**
- * Minimum time between rebuilds triggered by an unknown-cell refusal. A rebuild
- * picks up an applet or group installed since the last build, but only once per
- * interval, so a flood of calls to a genuinely-unknown cell cannot drive an
- * unthrottled listApps per call. Legitimate new cells do not depend on this
- * timing: the install IPC handlers call invalidate(), which drops the cache so
- * the next call rebuilds regardless of the interval. The residual case is a cell
- * that appears with no main-process signal (an applet cloning its own DNA over
- * the app websocket) and is called within the interval of the last build.
+ * Minimum time between rebuild *attempts* (see maybeRebuild). A rebuild picks up
+ * an applet or group installed since the last build, but only once per interval,
+ * so a flood of unknown-cell calls — or a rejecting listApps during an outage —
+ * cannot drive an unthrottled round-trip per call. Legitimate new cells do not
+ * depend on this timing: the install IPC handlers call invalidate(), which drops
+ * the cache and resets the attempt clock so the next call rebuilds regardless of
+ * the interval. The residual case is a cell that appears with no main-process
+ * signal (an applet cloning its own DNA over the app websocket) and is called
+ * within the interval of the last build.
  */
 const MIN_UNKNOWN_REBUILD_INTERVAL_MS = 1_000;
 
@@ -62,6 +63,9 @@ export class AppletZomeCallAuthorizer {
     this.cache = undefined;
     this.generation += 1;
     this.rebuildInFlight = undefined;
+    // Allow the next authorize() to rebuild immediately rather than being
+    // rate-limited against the previous attempt.
+    this.lastAttemptAt = -Infinity;
   }
 
   async authorize(
@@ -78,29 +82,40 @@ export class AppletZomeCallAuthorizer {
       });
 
     if (!this.cache) {
-      // Cold start (or just invalidated): must block until we know the cells.
-      await this.rebuild();
-    } else if (
-      this.clock() - this.cache.builtAt > CACHE_MAX_AGE_MS &&
-      this.clock() - this.lastAttemptAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS
-    ) {
-      // Stale-while-revalidate: keep serving the current set and refresh in the
-      // background, at most once per interval. The .catch keeps a rejected
-      // listApps (socket reconnecting, conductor restarting) from becoming an
-      // unhandled rejection, a signing outage, or one doomed round-trip per call.
-      void this.rebuild().catch(() => {});
+      // Cold start (or just invalidated): block until we know the cells. If the
+      // rebuild fails (or is rate-limited during an outage) the cache stays
+      // undefined and we fail closed below.
+      await this.maybeRebuild();
+    } else if (this.clock() - this.cache.builtAt > CACHE_MAX_AGE_MS) {
+      // Stale-while-revalidate: keep serving the current set, refresh in the
+      // background.
+      void this.maybeRebuild();
     }
 
     let decision = evaluate();
-    if (
-      !decision.sign &&
-      decision.reason === 'unknown-cell' &&
-      (!this.cache || this.clock() - this.cache.builtAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS)
-    ) {
-      await this.rebuild();
+    if (!decision.sign && decision.reason === 'unknown-cell') {
+      // Refresh once to pick up a cell installed/cloned since the last build.
+      await this.maybeRebuild();
       decision = evaluate();
     }
     return decision;
+  }
+
+  /**
+   * Rebuild the cache, but at most one *attempt* per interval (whether it
+   * succeeds or fails), and never propagating a failure. This is the one rate
+   * limit and error boundary for all three authorize() call sites: a flood of
+   * unknown-cell calls, or a rejecting listApps during an admin-socket outage,
+   * fires at most one round-trip per interval, and callers always fall through to
+   * a fail-closed decision rather than throwing a raw socket error at the applet.
+   * Concurrent callers coalesce onto the in-flight rebuild.
+   */
+  private maybeRebuild(): Promise<void> {
+    if (this.rebuildInFlight) return this.rebuildInFlight.catch(() => {});
+    if (this.clock() - this.lastAttemptAt <= MIN_UNKNOWN_REBUILD_INTERVAL_MS) {
+      return Promise.resolve();
+    }
+    return this.rebuild().catch(() => {});
   }
 
   /**
@@ -122,6 +137,11 @@ export class AppletZomeCallAuthorizer {
     const generationAtStart = this.generation;
     const p = (async () => {
       try {
+        // Suspend before calling listApps so the `this.rebuildInFlight = p` below
+        // always runs before the body can complete — otherwise a *synchronously*
+        // throwing listApps would leave a rejected promise stored in
+        // rebuildInFlight that every future rebuild() short-circuits onto forever.
+        await Promise.resolve();
         const apps = await this.listApps();
         // An invalidate() during the listApps round-trip means this snapshot may
         // already be stale (e.g. the app it would grant was just uninstalled), so
--- a/src/main/appletZomeCallAuthorizer.ts
+++ b/src/main/appletZomeCallAuthorizer.ts
@@ -1,0 +1,105 @@
+import { AppInfo, CallZomeRequest, CellId, CellInfo, encodeHashToBase64 } from '@holochain/client';
+import { appletIdFromAppId, getCellId } from '@theweave/utils';
+import { decideZomeCallSignable, type ZomeCallSigningDecision } from './zomeCallSigningPolicy';
+
+/**
+ * Resolves which cells each installed applet is allowed to have zome calls signed
+ * for, and decides individual sign requests against that set (via the pure
+ * decideZomeCallSignable policy).
+ *
+ * Resolving the allowed cells is a listApps round-trip and signing is on the hot
+ * path, so the result is cached. The cache is rebuilt lazily: a request that the
+ * cached view would refuse for an *unknown* cell triggers one refresh before the
+ * refusal, so a just-installed applet or group is picked up with no false
+ * rejection; concurrent refreshes coalesce into a single listApps. A known group
+ * cell refused for a non-profiles zome is not re-checked (a rebuild cannot change
+ * that verdict). Callers invalidate() on uninstall so the cache never keeps
+ * granting for an app that is gone.
+ */
+export class AppletZomeCallAuthorizer {
+  private cache: { ownByAppletId: Map<string, Set<string>>; groupCells: Set<string> } | undefined;
+  private rebuildInFlight: Promise<void> | undefined;
+
+  constructor(private listApps: () => Promise<AppInfo[]>) {}
+
+  /** Drop the cached cell sets so the next authorize() re-resolves from live app info. */
+  invalidate(): void {
+    this.cache = undefined;
+  }
+
+  async authorize(
+    callerAppletIds: string[],
+    zomeCall: CallZomeRequest,
+  ): Promise<ZomeCallSigningDecision> {
+    if (!zomeCall.cell_id) return { sign: false, reason: 'unknown-cell' };
+    const evaluate = (): ZomeCallSigningDecision =>
+      decideZomeCallSignable({
+        cellIdB64: cellIdKey(zomeCall.cell_id!),
+        zomeName: zomeCall.zome_name,
+        appletOwnCellIdsB64: this.ownCellsForCallers(callerAppletIds),
+        groupCellIdsB64: this.cache?.groupCells ?? new Set(),
+      });
+
+    if (!this.cache) await this.rebuild();
+    let decision = evaluate();
+    if (!decision.sign && decision.reason === 'unknown-cell') {
+      await this.rebuild();
+      decision = evaluate();
+    }
+    return decision;
+  }
+
+  /**
+   * The union of own-cells across the caller's applet ids. A plain applet caller
+   * passes one id; a cross-group view passes every applet of its tool, so it may
+   * legitimately sign for any of them (but still only their own cells).
+   */
+  private ownCellsForCallers(callerAppletIds: string[]): Set<string> {
+    const union = new Set<string>();
+    for (const id of callerAppletIds) {
+      for (const key of this.cache?.ownByAppletId.get(id) ?? []) union.add(key);
+    }
+    return union;
+  }
+
+  private rebuild(): Promise<void> {
+    if (this.rebuildInFlight) return this.rebuildInFlight;
+    this.rebuildInFlight = (async () => {
+      try {
+        const apps = await this.listApps();
+        const ownByAppletId = new Map<string, Set<string>>();
+        const groupCells = new Set<string>();
+        for (const app of apps) {
+          const appId = app.installed_app_id;
+          if (appId.startsWith('applet#')) {
+            ownByAppletId.set(appletIdFromAppId(appId), new Set(cellIdsOfApp(app)));
+          } else if (appId.startsWith('group#')) {
+            // Only the `group` role cell hosts the profiles zome an applet may call.
+            for (const key of cellIdsOfApp(app, 'group')) groupCells.add(key);
+          }
+        }
+        this.cache = { ownByAppletId, groupCells };
+      } finally {
+        this.rebuildInFlight = undefined;
+      }
+    })();
+    return this.rebuildInFlight;
+  }
+}
+
+/** Stable string key for a CellId ([DnaHash, AgentPubKey]) so it can go in a Set. */
+function cellIdKey(cellId: CellId): string {
+  return `${encodeHashToBase64(cellId[0])}|${encodeHashToBase64(cellId[1])}`;
+}
+
+function cellIdsOfApp(appInfo: AppInfo, roleFilter?: string): string[] {
+  const keys: string[] = [];
+  for (const [roleName, cellInfos] of Object.entries(appInfo.cell_info)) {
+    if (roleFilter && roleName !== roleFilter) continue;
+    for (const cellInfo of cellInfos as CellInfo[]) {
+      const cellId = getCellId(cellInfo);
+      if (cellId) keys.push(cellIdKey(cellId));
+    }
+  }
+  return keys;
+}
--- a/src/main/appletZomeCallAuthorizer.ts
+++ b/src/main/appletZomeCallAuthorizer.ts
@@ -47,6 +47,10 @@ export class AppletZomeCallAuthorizer {
     | undefined;
   private generation = 0;
   private rebuildInFlight: Promise<void> | undefined;
+  // When the last rebuild was *attempted* (vs `cache.builtAt` = last success).
+  // Rate-limits background refreshes so a rejecting listApps (admin socket
+  // outage) can't fire one doomed round-trip per authorize call.
+  private lastAttemptAt = -Infinity;
 
   constructor(
     private listApps: () => Promise<AppInfo[]>,
@@ -76,11 +80,14 @@ export class AppletZomeCallAuthorizer {
     if (!this.cache) {
       // Cold start (or just invalidated): must block until we know the cells.
       await this.rebuild();
-    } else if (this.clock() - this.cache.builtAt > CACHE_MAX_AGE_MS) {
+    } else if (
+      this.clock() - this.cache.builtAt > CACHE_MAX_AGE_MS &&
+      this.clock() - this.lastAttemptAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS
+    ) {
       // Stale-while-revalidate: keep serving the current set and refresh in the
-      // background. The .catch keeps a rejected listApps (socket reconnecting,
-      // conductor restarting) from becoming an unhandled rejection or a signing
-      // outage.
+      // background, at most once per interval. The .catch keeps a rejected
+      // listApps (socket reconnecting, conductor restarting) from becoming an
+      // unhandled rejection, a signing outage, or one doomed round-trip per call.
       void this.rebuild().catch(() => {});
     }
 
@@ -111,6 +118,7 @@ export class AppletZomeCallAuthorizer {
 
   private rebuild(): Promise<void> {
     if (this.rebuildInFlight) return this.rebuildInFlight;
+    this.lastAttemptAt = this.clock();
     const generationAtStart = this.generation;
     const p = (async () => {
       try {
--- a/src/main/appletZomeCallAuthorizer.ts
+++ b/src/main/appletZomeCallAuthorizer.ts
@@ -1,6 +1,27 @@
 import { AppInfo, CallZomeRequest, CellId, CellInfo, encodeHashToBase64 } from '@holochain/client';
 import { appletIdFromAppId, getCellId } from '@theweave/utils';
+import type { AppletId } from '@theweave/api';
 import { decideZomeCallSignable, type ZomeCallSigningDecision } from './zomeCallSigningPolicy';
+
+/**
+ * How long a resolved cell set may be trusted before it is rebuilt from live app
+ * info. This is the self-heal window for an uninstall that never called
+ * invalidate() — notably the renderer's leaveGroup, which uninstalls apps
+ * through its own admin websocket and so bypasses the main-process invalidate
+ * sites. Within this window a torn-down applet's cells can still be granted, but
+ * there is no iframe left to request signing for them.
+ */
+const CACHE_MAX_AGE_MS = 60_000;
+
+/**
+ * Minimum time between rebuilds triggered by an unknown-cell refusal. A rebuild
+ * picks up an applet or group installed since the last build (so a legitimate
+ * just-installed cell is not falsely refused), but only once per interval, so a
+ * flood of calls to a genuinely-unknown cell cannot drive an unthrottled
+ * listApps per call. The interval is well below the time it takes a
+ * newly-installed applet's iframe to mount and make its first call.
+ */
+const MIN_UNKNOWN_REBUILD_INTERVAL_MS = 1_000;
 
 /**
  * Resolves which cells each installed applet is allowed to have zome calls signed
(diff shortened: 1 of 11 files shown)


zippy:
Thanks — strong review. Addressed the two clusters and the small items in `8708d3c0` + `1c90c2fe`.

**CI red is pre-existing, not this PR.** The failing spec is smoke #03 (join-group), and it's red on `main-0.7` itself (since the invite-code feature added a second `.copy-link-input`, so the helper's locator matches two elements). My diff touches none of the invite/join UI. Fixed separately in #234 (a one-line `.first()`); merging that un-reds the smoke suite for this branch too. `unit-and-typecheck` (the required check) is green.

**Conductor launch (`8708d3c0`):**
- Kill the conductor in `finishErr` (and via it, on the 300s timeout) so a failed/slow launch can't leave an orphan holding the admin port + SQLite locks.
- Post-startup crash now emits on `MOSS_ERROR` (which `logs.ts` subscribes to) instead of `HOLOCHAIN_FATAL_PANIC` (zero subscribers → was dropped).
- Added `HolochainManager.shutdown()` (sets `shuttingDown`, then kills); the four deliberate-stop sites (factory reset, two restarts, quit) call it, and the post-startup exit listener returns early when `shuttingDown` — so a clean quit no longer reports a spurious crash.

**Authorizer cache (`1c90c2fe`):**
- Generation counter: an `invalidate()` during an in-flight `listApps` discards that snapshot instead of committing it.
- Max-age self-heal (60s): covers the `leaveGroup` bypass (renderer uninstalls via its own admin ws) — the cache rebuilds when stale, so a missed invalidate can't grant indefinitely. Docstring corrected to say this rather than overclaim.
- Throttled unknown-cell rebuild (≥1s): bounds `listApps` under a flood of refused calls, while still picking up a genuinely just-installed cell (its iframe mounts well after the interval — preserves the #231 finding-E behavior).
- `AppletId[]` restored (was widened to `string[]`); injected clock; 14 table tests incl. in-flight-invalidate, self-heal, throttle, just-installed.
- Dropped the prior-behavior narration from the applet-host `ready`/`search` comment (rule 6).

**Deferred, with reasoning:**
- **Hoisting the URL constants into `shared/utils`** — agreed it's the root de-dup, but this PR already repairs the immediate drift by updating both copies; hoisting touches wdocker's deps and the dev-mode accident-check that reads them, so it's a cleaner standalone change. Noting as a follow-up.
- **Wire-contract guard vs stale `libs/api/dist`** — the same staleness affects `typecheck` generally, and CI builds libs first, so it's consistent there; the error when it fires (`Type 'false' does not satisfy 'true'` on `WireContractInSync`) is the standard type-assert shape. Left as-is; happy to add a friendlier branded message if you'd prefer.

src/main/appletZomeCallAuthorizer.ts
```diff
@@ -0,0 +1,143 @@
...
+        } else if (appId.startsWith('group#')) {
+          // Only the `group` role cell hosts the profiles zome an applet may call.
+          for (const key of cellIdsOfApp(app, 'group')) groupCells.add(key);
+        }
+      }
+      this.cache = { ownByAppletId, groupCells };
+      this.builtAt = this.clock();
+    })();
+    this.rebuildInFlight = p;
+    void p.finally(() => {
```
zippy: `void p.finally(...)` creates a second promise that adopts `p`'s rejection with no handler attached. `p` itself is awaited by `authorize()`, so a `listApps` rejection is handled there — but the `.finally()`-derived promise rejects unhandled, and Electron's main process runs with Node's default `--unhandled-rejections=throw` (no `process.on('unhandledRejection')` anywhere in `src/`).

Failure path: the conductor dies after startup (the case the new exit listener reports), the admin websocket closes, the next applet zome call reaches `rebuild()`, `listApps` rejects — one failed signing request becomes a main-process crash.

The pre-refactor inline version used `try/finally` inside the async IIFE — one promise, always awaited. Restoring that shape fixes it:

```ts
const p = (async () => {
  try {
    const apps = await this.listApps();
    if (this.rebuildInFlight !== p) return; // superseded by invalidate()
    ...
    this.cache = { ownByAppletId, groupCells };
    this.builtAt = this.clock();
  } finally {
    if (this.rebuildInFlight === p) this.rebuildInFlight = undefined;
  }
})();
this.rebuildInFlight = p;
return p;
```

Note the `this.rebuildInFlight !== p` check is equivalent to the generation counter (`invalidate()` clears `rebuildInFlight`), so `generation` can be deleted along the way — one staleness protocol instead of two.

src/main/appletZomeCallAuthorizer.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  private rebuild(): Promise<void> {
+    if (this.rebuildInFlight) return this.rebuildInFlight;
+    const generationAtStart = this.generation;
+    const p = (async () => {
+      const apps = await this.listApps();
+      // An invalidate() during the listApps round-trip means this snapshot may
+      // already be stale (e.g. the app it would grant was just uninstalled), so
+      // discard it rather than restore it.
+      if (generationAtStart !== this.generation) return;
```
zippy: A discarded rebuild strands the authorizer in a state where a legitimate call is refused. `invalidate()` clears `cache` but not `builtAt`, and this early return commits neither. Sequence:

1. Cache commits at T (builtAt = T).
2. At T+300ms a call for a not-yet-cached cell starts a rebuild.
3. `invalidate()` lands during the listApps round-trip.
4. The snapshot is discarded here — `cache` is `undefined`, `builtAt` is still T.
5. `evaluate()` → `unknown-cell`; the throttle asks `clock() - builtAt > 1s`, which is false, so the recovery rebuild is skipped and `authorize` returns `{sign: false, reason: 'unknown-cell'}` for a cell the caller owns. Every caller coalesced onto that rebuild is refused too.

Either fix works: reset `builtAt = 0` in `invalidate()`, or move `builtAt` into the cache object so "no cache but fresh builtAt" is unrepresentable and the throttle reads `!this.cache || age > MIN_UNKNOWN_REBUILD_INTERVAL_MS`.

src/main/appletZomeCallAuthorizer.ts
```diff
@@ -0,0 +1,143 @@
...
+        appletOwnCellIdsB64: this.ownCellsForCallers(callerAppletIds),
+        groupCellIdsB64: this.cache?.groupCells ?? new Set(),
+      });
+
+    if (!this.cache || this.clock() - this.builtAt > CACHE_MAX_AGE_MS) await this.rebuild();
+    let decision = evaluate();
+    if (
+      !decision.sign &&
+      decision.reason === 'unknown-cell' &&
+      this.clock() - this.builtAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS
```
zippy: The 1s throttle drops the guarantee the pre-refactor code documented — "a miss always refreshes once before the call is refused, so a just-installed applet or group is picked up immediately with no false rejection" — and nothing on the install side compensates: neither install handler calls `invalidate()`, and clone cells are created renderer-side (`applet-host.ts` → `createCloneCell` over the app websocket), so main never hears about them at all.

Any two cells appearing within 1s of the last build mean the second one's first call is hard-refused ("Refusing to sign zome call … unknown-cell") with no retry anywhere in the path. Concrete cases: an applet creating two clone cells back-to-back and calling into each; two applets installed in quick succession during a group join whose iframes mount within a second. The header comment reasons about the applet-install mount time but not clones.

Suggest: skip the throttle when `!this.cache`, and add `invalidate()` (or a seed from the `AppInfo` already in hand) to the install handlers, mirroring the uninstall side. The throttle then only bounds the genuinely-unknown-cell flood it exists for. (Related: launch already has `listApps` results in hand twice before any applet mounts — seeding the authorizer from `postLaunchApps` would remove the cold-start rebuild from the first burst of zome calls entirely.)

src/main/appletZomeCallAuthorizer.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (!zomeCall.cell_id) return { sign: false, reason: 'unknown-cell' };
+    const evaluate = (): ZomeCallSigningDecision =>
+      decideZomeCallSignable({
+        cellIdB64: cellIdKey(zomeCall.cell_id!),
+        zomeName: zomeCall.zome_name,
+        appletOwnCellIdsB64: this.ownCellsForCallers(callerAppletIds),
+        groupCellIdsB64: this.cache?.groupCells ?? new Set(),
+      });
+
+    if (!this.cache || this.clock() - this.builtAt > CACHE_MAX_AGE_MS) await this.rebuild();
```
zippy: Doing the TTL rebuild synchronously on the signing path has two costs:

1. **No stale fallback.** Once the cache is >60s old, every zome call awaits `rebuild()`, and a rejected `listApps` (admin websocket reconnect, conductor busy or restarting) propagates — a total signing outage across all applets until the socket recovers. The pre-TTL warm cache kept working through exactly this kind of blip, and the 60s window already tolerates staleness by design, so serving the stale cache while a rebuild runs or fails costs nothing in the security model.
2. **Latency.** An applet that polls slower than every 60s pays a full `listApps` round-trip on every single call (its cache is always expired when it arrives), and in busy sessions whichever call lands first each minute stalls behind the rebuild.

Stale-while-revalidate fixes both: on expiry, fire `void this.rebuild()` (with a `.catch` — see the unhandled-rejection comment) and evaluate against the existing cache; only `await` when `this.cache` is undefined.

src/main/holochainManager.ts
```diff
@@ -161,14 +173,60 @@ export class HolochainManager {
...
+        if (settled) return;
+        settled = true;
+        clearTimeout(timeout);
+        resolve(manager);
+      };
+      const finishErr = (message: string) => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timeout);
+        conductorHandle.kill();
```
zippy, line 222: Killing here closes the orphan-process hole from the last review, but two paths that reach `finishErr` kill a conductor that is healthy:

- **The 300s timeout can't tell hung from slow.** First launch after a Holochain version bump — DB migration plus cold WASM compile on a slow disk — can exceed it; the timer then SIGTERMs mid-migration, and the retry restarts the same work, possibly against a half-migrated DB. That's a repeat-failure loop the old hang couldn't produce. Consider resetting the timer whenever recognizable startup progress appears on stdout, rather than one absolute deadline.
- **`Holochain conductor ready but failed to connect` fires after `Conductor ready.`** — the conductor is fully up and the only fault was the admin-websocket connect — and now kills it. Probably still the right call (nothing else owns the process at that point), but worth being deliberate that this branch is "kill a healthy conductor because we couldn't reach it".

src/main/holochainManager.ts
```diff
@@ -197,20 +255,29 @@ export class HolochainManager {
...
+              installedApps,
+              version,
             );
+            // Report a crash after a successful startup, but not a deliberate
+            // stop (quit / restart / factory reset, which set `shuttingDown`).
+            // MOSS_ERROR is used because it has a subscriber (logs.ts); the
+            // fatal-panic channel has none, so it would drop the report.
+            conductorHandle.on('exit', (code, signal) => {
+              if (manager.shuttingDown) return;
+              weEmitter.emitMossError(
```
zippy, line 320: MOSS_ERROR's only subscriber is `logs.ts` (console + winston file), so a post-startup crash now produces a log line — a real improvement over the subscriber-less fatal-panic channel — but the user-facing outcome is unchanged: the UI keeps rendering against a dead socket and every zome call fails opaquely. Deferring actual renderer surfacing is reasonable; a TODO naming the missing renderer notification would keep the gap visible rather than letting the comment read as "handled".

Also note the interaction with the authorizer finding: with `void p.finally(...)` as-is, the first zome call after this exact crash can take down the main process before anyone reads the log.

zippy (commented):
Re-review of the follow-up commits (8708d3c0 + 1c90c2fe). The previous round's findings are all verifiably addressed: `finishErr` kills the conductor, the four kill sites go through `shutdown()`, the crash report moved to a subscribed channel, the generation counter closes the invalidate race it was aimed at, the 60s TTL covers the renderer-side uninstall bypass (and I agree with the deferral there — the uninstalled cells no longer exist in the conductor, so the ≤60s stale grant has no reachable consumer), `AppletId[]` is restored, and the drift test still passes (78/78, no dead, no orphans).

The new mechanisms introduce four issues worth fixing before merge — see the inline comments. The one hard bug is the `void p.finally(...)` unhandled rejection, which can take down the main process on the exact conductor-crash path this PR now reports. The other three inline comments (discarded-rebuild refusal, the 1s throttle dropping the always-rebuild-before-refuse guarantee, the blocking TTL rebuild with no stale fallback) are all in the authorizer's freshness machinery and share a fix family.

Smaller items, no inline anchor:

- **`generation` is redundant.** `invalidate()` already clears `rebuildInFlight`, so `if (this.rebuildInFlight !== p) return;` after the `await` is equivalent to the generation check, and a `try/finally` inside the IIFE replaces the detached `.finally()` chain (fixing the unhandled rejection at the same time). One staleness protocol instead of two.
- **`'Conductor ready.'` branch is not guarded by `settled`.** A second matching stdout line opens a second AdminWebsocket and attaches a second post-startup exit listener closed over a different `manager` — one that never gets `shuttingDown`, so a normal quit would log a spurious crash. `if (settled) return;` at the top of the branch.
- **Launch rejection can orphan lair.** `launch()` spawns lair before `HolochainManager.launch()`; on rejection the destructuring at the call site never runs, `LAIR_HANDLE` stays undefined, and the quit handler has nothing to kill — each splash-screen retry spawns another lair against the same keystore dir. Worth killing lair in `launch()`'s failure path, symmetrical with the conductor kill this PR added.
- **`ipcMain.handle('exit')` → `app.exit(0)` bypasses `shutdown()` entirely** (splash-screen quit button): conductor and lair survive holding the port and SQLite locks. Pre-existing, but the new `shuttingDown` convention makes it worth a look — `app.quit()` instead of `app.exit(0)` would run the quit handler.
- **Rule-6 comments**: the drift-test docblock still narrates the pre-PR state ("five preload channels had no handler"), and the new comments at holochainManager.ts:213-215 ("handled by the listener below") and 270-271 ("MOSS_ERROR is used because...") are positional/contrast comments that will rot when either half moves.
- **On the stale-dist deferral**: CI is covered, but plain `yarn build` runs `typecheck` without `build:libs`, so locally both new compile-time guards can pass against a stale `libs/api/dist`. A `paths` mapping to `libs/api/src` in tsconfig (or making typecheck depend on build:api) closes it cheaply.
- **validationSchemas nits**: the schema union has duplicated members — `asset-to-pocket` and `unsubscribe-from-asset-store` each appear twice, and the second `unsubscribe-from-asset-store` (~line 526) looks like a copy-paste that was meant to be a different message; the doc comment also says `_WireContractInSync` while the export is `WireContractInSync`.

zippy:
Thanks — all addressed in `3c3bd422` + `b7508507`.

**The hard bug — unhandled rejection (authorizer:121).** Fixed: `rebuild()` is now a single awaited promise with `try/finally` inside the IIFE, no detached `.finally()`. A rejected `listApps` (the conductor-crash path) propagates only to the awaiting caller. I kept the `generation` counter rather than the `rebuildInFlight !== p` identity check — the latter reads `p` inside its own initializer, which TS flags; generation avoids that with the same semantics.

**Authorizer freshness (the shared fix family):**
- **Discarded-rebuild strand (105):** `builtAt` moved into the cache object, so "no cache but fresh builtAt" is unrepresentable — after a discard, `!cache` forces a rebuild instead of a false refusal. New test pins it.
- **Throttle dropping the guarantee (76):** `invalidate()` is now wired into the install handlers (each `enableApp`), so a just-installed cell is grantable at once regardless of the throttle. Residual, documented in the header: a clone cell created renderer-side and called within 1s of the last build — no main-process signal exists for that path.
- **Blocking TTL rebuild (71):** now stale-while-revalidate — past the max age the warm cache is served and the refresh runs in the background with a `.catch`, so a rejected `listApps` during a socket reconnect no longer blocks or fails all signing. Tests for both the serve-stale and the reject-doesn't-block cases.

**Conductor / lifecycle (smaller items):**
- Ready-timeout is now an **inactivity** timer (reset on any stdout/stderr line), so a slow-but-progressing first-run migration isn't SIGTERMed.
- `'Conductor ready.'` branch guarded by `settled` — no second websocket / second exit listener.
- **Lair orphan:** killed in `launch()`'s failure paths (HolochainManager and WeRustHandler), symmetric with the conductor kill.
- **`exit` handler:** kept `app.exit(0)` (immediate) but now stops the conductor + lair first, so they don't survive holding the port/keystore. (Chose this over `app.quit()` to avoid changing the exit semantics of a pre-existing handler.)
- MOSS_ERROR crash report keeps its log subscriber; added a **TODO** naming the missing renderer-facing notification. Comments de-narrated (rule 6).

**validationSchemas nits:** the duplicate `unsubscribe-from-asset-store` was pure cruft, not a missing message — `subscribe-to-asset-store` is present separately, which is why `WireContractInSync` passed. Removed both duplicates + the `asset-to-pocket` one; fixed the `_WireContractInSync` docblock name.

**Deferred:** the `paths → libs/api/src` mapping. It changes module resolution for the whole renderer typecheck (src vs the published dist shape), which is a broader build-config change than a review fix should carry; CI already builds libs before typecheck. Happy to do it as its own change if you want the local `yarn build` gap closed.

Verified: `typecheck` + unit (18 authorizer tests incl. the new stale-serve / reject / strand cases) green; e2e #12 still passes (it installs an applet then immediately calls it, exercising the new invalidate-on-install path).

zippy (commented):
Re-review of 3c3bd422 + b7508507. All of round two's findings are verifiably addressed: `rebuild()` is a single awaited promise with `try/finally` (no detached chain), `builtAt` lives inside the cache and `!this.cache` bypasses the throttle (the strand test covers exactly the discarded-cold-rebuild case), stale-while-revalidate keeps a rejected `listApps` from causing a signing outage, invalidate is wired into the install handlers, lair is killed in the covered launch failure paths, the exit IPC stops the stack before `app.exit(0)`, and the inactivity timer removes the kill-a-slow-migration hazard. Keeping the generation counter over the promise-identity check is a defensible call given the TS use-before-assign flag. And you were right about the schema duplicates — I checked, `subscribe-to-asset-store` is present, so my "meant to be a different message" guess was wrong; they were pure cruft.

This round found one concrete miss and three narrow gaps in the new machinery — see inline. Ranked: the group-import loop is the one real invalidate hole (and the fact that a 9th call site missed the convention is itself the argument for making the pairing structural); the inactivity timer's two issues; the ready-branch guard TOCTOU; the launch cleanup coverage gap. The clone-cell window is the one residual regression vs pre-PR behavior, noted inline with options.

Smaller items, no anchor:

- **`install-app` handler (index.ts:1553)**: `HOLOCHAIN_MANAGER.installApp(filePath, appId, networkSeed)` takes an arbitrary appId — if anything can pass an `applet#` id through it, it needs the same invalidate. Worth a quick check.
- **`shutdown()` doesn't await process exit**: factory reset calls it and then immediately deletes the conductor SQLite databases and keystore out from under a process that may still be dying — on Windows that's an EBUSY/EPERM path where `factoryReset()` rejects after windows are closed and the relaunch never happens. Pre-existing ordering, but `shutdown()` is now the orderly-stop primitive and could return a promise that resolves on the `exit` event.
- **`processHandle` is still public**, so `shutdown()` is a convention — a future direct `kill()` reintroduces the spurious-crash report. Making it private with a `get pid()` accessor (the one remaining external read, index.ts:2792) turns the convention into a compile error. Same eventually for the raw `LAIR_HANDLE.kill()` sites and `LOCAL_SERVICES_HANDLE`, which the `exit` IPC teardown does not stop — a `stopSubprocesses()` shared by exit/quit/reset/relaunch would end the enumerate-at-every-site pattern.
- **Drift test roots at `process.cwd()`** — running vitest from anywhere but the repo root fails at collection with ENOENT instead of a named test failure; `import.meta.dirname`-relative is cwd-independent.
- **Still open from round two**: whether the live 0.6 `bootstrap.moss.social` should be re-added to the production-URL lists (it's currently in neither copy, so the cli.ts dev-mode guard lets it through), alongside the deferred hoist to shared/utils.

None of this changes the merge order recommendation (#234 first, then this). The group-import invalidate, the `if (!settled)` timer guard, and the ready-branch flag are one-to-three-line fixes worth taking before merge; the rest can trail.

src/main/index.ts
```diff
@@ -2473,6 +2394,7 @@ if (!RUNNING_WITH_COMMAND) {
         console.log('Determined appId for group: ', appId);
         if (apps.map((appInfo) => appInfo.installed_app_id).includes(appId)) {
           await HOLOCHAIN_MANAGER!.adminWebsocket.enableApp({ installed_app_id: appId });
+          appletZomeCallAuthorizer.invalidate();
```
zippy: One install/enable site was missed: the group-import loop (`runGroupsImport`, ~line 2310) does `installApp` + `enableApp` per imported applet with no `appletZomeCallAuthorizer.invalidate()` — the group-happ install in that same handler has one, as do the other seven sites including this one. (Can't anchor there since the loop is untouched by this PR.) The import loop installs several applets back to back while live iframes are making sign calls, so an unknown-cell rebuild triggered mid-loop leaves a <1s-old cache that omits the next imported applet, and its first zome call is refused ("unknown-cell") inside the throttle window.

Beyond the one-line fix, this miss is the argument for making the pairing structural instead of conventional: a ninth call site missed it in the same PR that established the convention. Two local helpers (`enableAppInvalidating` / `uninstallAppInvalidating`), or folding the invalidate into `HolochainManager` wrappers, gives the next contributor compiler help and shrinks all nine sites.

src/main/holochainManager.ts
```diff
@@ -161,18 +173,75 @@ export class HolochainManager {
...
+      // distinguish a deliberate shutdown from a crash.
+      conductorHandle.on('exit', (code, signal) => {
+        if (settled) return;
+        finishErr(
+          `Holochain conductor exited during startup (code ${code}, signal ${signal}) before becoming ready. Check the logs for details (Help > Open Logs).`,
+        );
+      });
+
       conductorHandle.stdout.pipe(split()).on('data', async (line: string) => {
+        armInactivityTimer();
```
zippy: Two issues with the re-arm placement:

1. **It never stops.** `armInactivityTimer()` runs on every stdout line (and stderr, line 298) for the process's whole lifetime — Holochain logs continuously, so that's a clearTimeout + 300s setTimeout per log line forever, a perpetually pending timer whose `finishErr` callback is a guaranteed no-op after settle, pinning the launch closure and keeping the event loop alive for any non-Electron consumer of `launch()`. `if (!settled) armInactivityTimer();` reduces it to a boolean check per line.

2. **Silence is not readiness.** The timer only fires when the conductor goes quiet, so a conductor that keeps logging but never prints `Conductor ready.` — a pre-ready retry loop, or a future Holochain that rewords the magic string — hangs the launch forever, which is the original bug this timer family exists to prevent. The absolute deadline you replaced had the opposite failure mode (killing slow-but-progressing migrations); the two compose: keep the inactivity timer as the common-case backstop and add a generous absolute cap (say 15–20 min) behind it, or treat a bounded-retry `AdminWebsocket.connect` poll as the structured readiness signal so the string match stops being the only path to resolve.

src/main/holochainManager.ts
```diff
@@ -161,18 +173,75 @@ export class HolochainManager {
...
         if (line.includes('could not be parsed, because it is not valid YAML')) {
-          reject(
+          finishErr(
             `Holochain failed to start up and crashed. Check the logs for details (Help > Open Logs).`,
           );
         }
         if (line.includes('Conductor ready.')) {
+          // Guard against a second matching line opening a second admin websocket
+          // and attaching a second post-startup exit listener.
+          if (settled) return;
```
zippy: This guard is checked before the `await`s but `settled` is only set in `finishOk` at the end of the connect sequence, so it doesn't prevent what the comment says it prevents: a second `Conductor ready.` line arriving while the first handler is still awaiting `AdminWebsocket.connect`/`listApps`/`attachAppInterface` passes the check and runs the whole block again — second admin websocket, possibly a second app interface, and a second post-startup exit listener closed over a manager that never receives `shuttingDown` (so a normal quit logs a spurious crash).

Fix is a flag set synchronously before the first `await`: `let readyHandled = false;` … `if (settled || readyHandled) return; readyHandled = true;`.

src/main/launch.ts
```diff
@@ -114,6 +117,10 @@ export async function launch(
   try {
     weRustHandler = await rustUtils.WeRustHandler.connect(lairUrl, password);
   } catch (e) {
+    // Both lair and the conductor are running now; kill both so a failed launch
+    // doesn't leave them orphaned holding the keystore and the admin port.
+    lairHandle.kill();
```
zippy, line 122: The kill-on-failure coverage stops here. Everything after this catch — the default-apps install loop and `devSetup()` — still throws without killing lair or the conductor, and because the throw happens before the destructuring assignment in the `launch` IPC handler, `LAIR_HANDLE`/`HOLOCHAIN_MANAGER` stay undefined and every later cleanup path skips them. Concretely: `yarn applet-dev` with a broken dev config (or a default-app webhapp that fails validation) leaves a live conductor holding the admin port and SQLite locks plus a live lair on the keystore — the exact orphan class these two catches were added to eliminate.

One try/catch around all the post-spawn stages whose catch tears down whatever exists so far (`lairHandle.kill(); holochainManager?.shutdown();`) covers today's stages and whatever gets added next, instead of a third copy of the cleanup pair per stage.

src/main/appletZomeCallAuthorizer.ts
```diff
@@ -0,0 +1,158 @@
...
+      // conductor restarting) from becoming an unhandled rejection or a signing
+      // outage.
+      void this.rebuild().catch(() => {});
+    }
+
+    let decision = evaluate();
+    if (
+      !decision.sign &&
+      decision.reason === 'unknown-cell' &&
+      (!this.cache || this.clock() - this.cache.builtAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS)
```
zippy: With installs now invalidating, the residual regression vs the pre-PR always-rebuild-on-miss is clone cells: `create-clone-cell` is handled entirely renderer-side over the app websocket, so main never hears about it, and a clone created within 1s of the last build gets its first zome call hard-refused ("unknown-cell") with nothing retrying. An applet that clones a DNA on startup — or creates two clones back to back — hits exactly this: the rebuild triggered by calling clone A leaves a <1s-old cache that lacks clone B.

Options, cheapest first: have the `create-clone-cell` case in applet-host also fire an IPC that calls `invalidate()` (symmetrical with the install handlers); or, when the throttle suppresses the rebuild, wait out the remainder of the interval and rebuild once instead of refusing (bounds listApps at the same 1/s while restoring the no-false-refusal guarantee); or accept the window and say so in the class doc, which currently only reasons about the applet-install mount time.

Smaller, same machinery: on the stale-refresh path above, a *failed* rebuild leaves `builtAt` old and `rebuildInFlight` cleared, so during an admin-socket outage every authorize fires another doomed background `listApps` — one per call, unbounded in count. A `lastAttemptAt` recorded on entry to `rebuild()` (reusing `MIN_UNKNOWN_REBUILD_INTERVAL_MS`) caps it at one retry per second.

zippy:
Round-3 addressed in `4efb170d` + `0dc3f87b`.

**The one real miss — group-import invalidate.** Fixed structurally, per your argument that a 9th missed site means convention isn't enough: all enable sites now go through a single `enableAppAndInvalidate` helper (import loop included), and the `install-app` handler — which enables inside `HolochainManager.installApp`, bypassing the helper — invalidates directly. Nine sites → one helper + two direct calls.

**Inactivity timer (both issues).** Re-armed only while unsettled (`if (!settled) armInactivityTimer()`), so it stops once the conductor is up and logging. And added a 20-min absolute cap behind it for the "logs forever but never ready" failure the silence timer can't see.

**Ready-branch TOCTOU.** `readyHandled` is now set synchronously before the connect `await`s, so a second `Conductor ready.` line mid-connect can't run the block twice.

**Launch cleanup coverage.** One try/catch now wraps all post-spawn stages (default-apps install + devSetup); its catch does `lairHandle.kill(); holochainManager.shutdown()`, so a broken dev config / failed default-app validation no longer orphans the stack past the two catches from last round.

**Failed-rebuild round-trips.** The stale-refresh path is gated on a `lastAttemptAt` timestamp, so a rejecting `listApps` during a socket outage fires at most one doomed round-trip per interval, not one per sign call.

**Drift test rooting.** Now resolves via `import.meta.url`, so running vitest from a subdir fails as a named assertion instead of ENOENT.

**install-app handler** — covered (invalidates after `HolochainManager.installApp`).

**Clone-cell window — documented, not code-fixed (option c).** The class doc names it explicitly (a cell that appears with no main-process signal — a renderer-side clone — called within 1s of the last build). I chose this over the IPC (option a, adds a channel to the surface I just made drift-tested) and the wait-out (option b, self-contained but reworks the throttle + several tests for a narrow window). Happy to take (a) or (b) if you'd rather close it than document it — your call.

**Deferred (noted, not in this PR):**
- **`shutdown()` doesn't await exit** → factory-reset DB deletion races a dying process (Windows EBUSY). Real, but making `shutdown()` return an exit-resolved promise changes its signature and every caller — better as its own change than folded into a review round.
- **`processHandle` private + a shared `stopSubprocesses()`** for exit/quit/reset/relaunch (ending the enumerate-at-every-site `LAIR_HANDLE.kill()` pattern). Agreed direction; it's a lifecycle refactor of its own.
- **Re-adding `bootstrap.moss.social` to the dev-mode guard list** — that's a #232 product call (should a 0.6 server still be guarded on the 0.7 line?) and touches cli.ts, which isn't in this PR.

Verified: `typecheck` + unit (145, incl. the new outage rate-limit test) green; e2e #12 passes — it installs an applet then immediately calls it, exercising the structural enable-invalidate path.

src/main/appletZomeCallAuthorizer.ts
```diff
@@ -0,0 +1,166 @@
...
+      // listApps (socket reconnecting, conductor restarting) from becoming an
+      // unhandled rejection, a signing outage, or one doomed round-trip per call.
+      void this.rebuild().catch(() => {});
+    }
+
+    let decision = evaluate();
+    if (
+      !decision.sign &&
+      decision.reason === 'unknown-cell' &&
+      (!this.cache || this.clock() - this.cache.builtAt > MIN_UNKNOWN_REBUILD_INTERVAL_MS)
```
zippy: Two gaps in this branch, both in the seam the `lastAttemptAt` fix was aimed at:

1. **It's throttled on the wrong timestamp.** The gate reads `cache.builtAt` (last *success*), which a failing rebuild never advances — so during an admin-socket outage every unknown-cell call passes the gate and fires another doomed `listApps`, exactly the one-round-trip-per-call the `lastAttemptAt` field was introduced to prevent (it currently guards only the stale-refresh branch above; the cold `!this.cache` branch is unthrottled too). The docblock's "a flood cannot drive an unthrottled listApps per call" doesn't hold on this path, and the outage test only exercises the stale-refresh branch.

2. **The `await this.rebuild()` has no catch**, so the raw `listApps` rejection propagates and `authorize()` throws instead of returning `{sign: false, reason: 'unknown-cell'}` — the applet gets an opaque socket error attributed to its own call.

One shape fixes both plus the cold branch: a private `maybeRebuild()` gated on `lastAttemptAt` for all three call sites, with the callers catching (or `maybeRebuild` swallowing) the failure and falling through to the refusal decision. One rule, one timestamp, one constant — `cache.builtAt` then feeds only the `CACHE_MAX_AGE_MS` staleness check.

src/main/appletZomeCallAuthorizer.ts
```diff
@@ -0,0 +1,166 @@
...
+            // Only the `group` role cell hosts the profiles zome an applet may call.
+            for (const key of cellIdsOfApp(app, 'group')) groupCells.add(key);
+          }
+        }
+        this.cache = { ownByAppletId, groupCells, builtAt: this.clock() };
+      } finally {
+        if (generationAtStart === this.generation) this.rebuildInFlight = undefined;
+      }
+    })();
+    this.rebuildInFlight = p;
```
zippy, line 166: Latent wedge in the assignment order: if `this.listApps()` throws *synchronously*, the async IIFE body runs to completion synchronously — the `try` throws before the first suspension, the `finally` runs (and its generation guard passes, clearing a `rebuildInFlight` that is still `undefined`), and only then does this line assign the already-rejected promise into `this.rebuildInFlight`. From then on every `rebuild()` short-circuits at the top and returns the same rejected promise forever; `lastAttemptAt` never advances and no new `listApps` is attempted — all signing is wedged until something happens to call `invalidate()`.

Not reachable today (the supplier `() => HOLOCHAIN_MANAGER!.adminWebsocket.listApps({})` only throws synchronously if the manager is gone, and the IPC handler guards that first), but it's one hostile refactor away. Cheapest fix: make the first statement of the `try` a suspension point — `const apps = await Promise.resolve().then(this.listApps);` — so the body cannot complete before the assignment; alternatively assign via `let` before invoking.

src/main/holochainManager.ts
```diff
@@ -197,30 +282,42 @@ export class HolochainManager {
...
+              installedApps,
+              version,
             );
+            // Report a crash after a successful startup, but not a deliberate
+            // stop (quit / restart / factory reset, which set `shuttingDown`).
+            // Emit on MOSS_ERROR so the crash reaches the log subscriber.
+            // TODO: also surface this to the renderer — the UI currently keeps
+            // rendering against a dead conductor and every zome call fails
+            // opaquely; there is no user-facing "conductor stopped" notification.
+            conductorHandle.on('exit', (code, signal) => {
```
zippy, line 318: The connect sequence never re-checks `settled` after its awaits, so a launch that already failed via the inactivity or absolute timer still finishes this block against the process `finishErr` just killed: `new HolochainManager(...)` is constructed, this exit listener is attached with `shuttingDown` false, and the SIGTERM's exit event logs `Holochain conductor exited unexpectedly after startup` — a launch reported as a startup timeout is also logged as a post-startup crash, and the AdminWebsocket stays open against a dead conductor (each splash retry leaks another, since `readyHandled` makes the path terminal).

An `if (settled) { adminWebsocket.client.close(); return; }` after the awaits (before constructing the manager) closes both the spurious crash report and the socket leak. The same close belongs in the `catch` → `finishErr` path, which currently kills the conductor but leaves the just-opened websocket connected.

src/main/holochainManager.ts
```diff
@@ -161,18 +173,91 @@ export class HolochainManager {
...
     conductorHandle.stdin.end();
 
+    // Two backstops for a conductor that never becomes ready. The inactivity
+    // timer measures silence, not total time, so a slow but progressing first-run
+    // migration + WASM compile (which keeps logging) is not killed. The absolute
+    // cap catches the opposite failure the inactivity timer cannot see — a
+    // conductor that keeps logging but never prints 'Conductor ready.' (a pre-ready
+    // retry loop, or a reworded magic string). Deterministic failures (crash / bad
+    // config / port conflict) are caught by the exit + error listeners.
+    const LAUNCH_INACTIVITY_TIMEOUT_MS = 300_000;
```
zippy: One risk worth weighing before this ships: the inactivity timer assumes a healthy slow startup keeps logging, but the conductor is launched with `RUST_LOG` defaulting to `warn` and the chattiest startup targets (wasmer cranelift, holochain_sqlite, kitsune discovery) explicitly demoted to `error` — so a first launch doing a long DB migration + WASM compile can legitimately emit nothing above the threshold for >5 minutes. If that happens, `finishErr` SIGTERMs the conductor mid-migration and every retry repeats the kill — on base the launch just took longer and succeeded.

I can't say with certainty how silent a real migration is at these log levels, so this may be theoretical — but the failure mode is bad enough (repeated kills mid-migration) to deserve either a much larger inactivity threshold (silence is only meant to catch true hangs; 10–15 min costs little), or the structural alternative: poll `AdminWebsocket.connect` against the known admin port as the readiness signal, with the absolute cap as the only deadline — that removes the dependency on both the log level and the `Conductor ready.` string.

src/main/index.ts
```diff
@@ -1247,6 +1171,10 @@ if (!RUNNING_WITH_COMMAND) {
 
   function registerIPCHandlers(notificationIcon: Electron.NativeImage) {
     ipcMain.handle('exit', () => {
+      // app.exit(0) skips the before-quit handler, so stop the subprocesses here
+      // or they survive as orphans holding the admin port and keystore.
+      if (HOLOCHAIN_MANAGER) HOLOCHAIN_MANAGER.shutdown();
```
zippy, line 1178: Two gaps against the `app.on('quit')` handler this stands in for, both latent today (the only renderer caller of `electronAPI.exit()` — splashscreen's `renderExitButton` — is currently never rendered):

1. `LOCAL_SERVICES_HANDLE` isn't killed and `.hc_local_services` isn't removed, so a dev-mode exit through this path orphans the kitsune2 bootstrap/relay child and leaves the stale services file for the next agent's `readLocalServices()`.
2. Both globals are only assigned after `await launch(...)` resolves, so an exit during the launch window (now up to 20 minutes) kills nothing — the in-flight lair and conductor are orphaned exactly as before the fix.

Given `app.on('quit')` already does the full teardown, the simplest correct form is `app.quit()` instead of `app.exit(0)` (the comment's stated reason for `app.exit` is that it skips before-quit — but that's the bug, not the feature, now that quit does the cleanup). If `app.exit` stays, this handler needs the local-services kill and, for the launch window, either a scoped handle set before launch resolves or acceptance of the gap in a comment.

zippy (commented):
Round-4 re-review at 0dc3f87b. Everything from round three is verifiably in: `enableAppAndInvalidate` at all six enable sites plus the import loop, direct invalidate in `install-app`, the settled-gated timer re-arm plus the 20-minute absolute cap, the synchronous `readyHandled` flag, the launch teardown try/catch, the `lastAttemptAt` rate limit, and the drift test rooted at the file. The clone-cell judgment call (document rather than plumb an IPC) is reasonable and I'm treating it as accepted; one sharpening for whenever it's revisited — `create-clone-cell` is a first-party host path in applet-host, so option (a) is one preload channel plus one `invalidate()` call, not a new surface.

This round's findings are all in the seams of the new machinery — see inline. Ranked: the unknown-cell branch missing the rate limit it motivated (and rejecting instead of refusing during an outage); a latent wedge from `rebuildInFlight` being assigned after a synchronous supplier throw has already run the `finally`; the ready-branch connect sequence never re-checking `settled` after its awaits; the inactivity timer's silent-healthy-startup risk; and the exit-handler teardown gaps (latent — the button isn't currently rendered).

Body-only items:

- **Post-crash recovery is still wedged**: the crash listener logs via MOSS_ERROR but nothing clears `HOLOCHAIN_MANAGER`, so `get-conductor-info` keeps returning the dead conductor's ports and moss-app's `if (!info) launch()` branch is unreachable — a window reload renders a normal-looking UI against a dead socket, and only a full app restart recovers. Clearing `HOLOCHAIN_MANAGER` (plus `appletZomeCallAuthorizer.invalidate()`) in that listener would make the existing renderer reconnect path do the work, and pairs naturally with the TODO on renderer surfacing.
- **`launch` handler reassigns `HOLOCHAIN_MANAGER` without invalidating the authorizer** — unreachable today (no in-process relaunch path), but one line next to the reassignment closes the class before someone adds a restart IPC.
- **`enableAppAndInvalidate` could be softer**: enable only widens the grantable set, so the hard `invalidate()` per enable forces the next live zome call to block on a full cold rebuild — during a batch join/import that's one blocking `listApps` per applet enabled. A background `void rebuild().catch(...)` (the stale-refresh shape) preserves correctness with none of the stalls; keep the hard invalidate for uninstall, where the set narrows and staleness is the security issue. Related docblock nit: "every enable site must invalidate" — `HolochainManager.installWebApp`/`installApp` enable internally and rely on their callers' invalidate (covered today; the install-app handler's comment acknowledges it), so the invariant actually enforced is "every *enable path reachable from IPC* invalidates".
- **Pre-existing, adjacent**: `conductorHandle.stdin.write(password)` runs with no `error` listener on the stdin stream, so a spawn failure (missing/bad binary) can surface as an uncaught EPIPE/ERR_STREAM_DESTROYED stream error that crashes main before the `'error'` listener converts it into the clean launch rejection it was added for. A no-op `conductorHandle.stdin.on('error', () => {})` (or writing only after `spawn` event) closes it.
- **Polish, take or leave**: the three launch.ts catch blocks repeat kill/shutdown+emit+throw with drifting contents (an `abortLaunch(stage, e)` helper collapses them); the try-wrap re-indents ~105 lines including the 35-line commented-out KanDo block whose brackets are commented but whose body is live — deleting it (git history keeps it) would shrink the diff to the two real lines; the authorizer test file hand-rolls four harness variants that a single `harness({apps, gated, fail})` would replace; the `>50` floor in the drift test tolerates a third of the surface silently vanishing from the scan (derive it from the handled set instead); comment nits per CLAUDE.md rule 6 (the `(bypassing enableAppAndInvalidate)` parenthetical, the applet-host tail quoting the removed runtime reject, the "instead of relying on each call site" clause); and the double blank line at electron-api.ts:400 has now survived three rounds.

On convergence: nothing here blocks merge. The three inline authorizer/lifecycle fixes (unify the rebuild throttle+catch, move the `rebuildInFlight` assignment, re-check `settled` after the connect awaits) are small and worth taking now; the rest — including the two altitude items I've raised before (magic-string readiness, HolochainManager as the lifecycle chokepoint) — are follow-up material, not reasons to keep this PR open a fifth round.

zippy:
Round-4 addressed in `12cc3ef1` + `f1807f5b`. The three ranked inline fixes plus the cheap seams; a couple deferred/pushed back with reasoning.

**Ranked inline — all fixed:**
1. **Authorizer rate-limit + error boundary.** All three call sites (cold, stale-refresh, unknown-cell) now go through one `maybeRebuild`: gated on `lastAttemptAt` (attempt time, not `cache.builtAt`), and swallowing failures so callers fall through to a fail-closed refusal instead of throwing a raw socket error. `invalidate()` resets the attempt clock so a fresh install still rebuilds at once. New test exercises the unknown-cell outage path the round-3 test missed.
2. **Sync-throw wedge.** A leading `await Promise.resolve()` in the rebuild IIFE makes `this.rebuildInFlight = p` always run before the body can complete, so a synchronously-throwing `listApps` can't leave a stuck rejected promise.
3. **Ready-branch re-check.** After the connect awaits it now `if (settled) { adminWebsocket.client.close(); return; }` before constructing the manager, and the catch closes the socket too — no spurious post-startup crash log, no leaked websocket when a timer fails the launch mid-connect.

**Cheap seams — done:**
- **Inactivity timer → 10 min** (silence only catches a true hang; the 20-min absolute cap is the real deadline), addressing the silent-slow-migration risk. The structural alternative (poll `AdminWebsocket.connect` as readiness) I'm leaving as the follow-up you flagged.
- **stdin `error` handler** so a spawn failure can't EPIPE-crash main before the process `error` listener converts it.
- **launch handler** invalidates the authorizer after (re)assigning `HOLOCHAIN_MANAGER`.
- **`exit` IPC** now also kills `LOCAL_SERVICES_HANDLE`; noted the launch-window gap it can't cover. Kept `app.exit(0)` (force-quit is the point of an exit button) rather than switch to `app.quit()` — reasonable either way; say the word if you'd prefer quit.
- Rule-6 comment nits + the electron-api double blank line (finally).

**Pushed back / deferred, with reasoning:**
- **Softening `enableAppAndInvalidate` to a background refresh.** Kept the hard `invalidate()`. A background refresh reintroduces a false-refuse window for a *just-enabled* cell called before the refresh lands (same class as the clone window we agreed to document), and risks flaking the install-then-call e2e; the cost it avoids is one cheap `listApps` per enable on the non-hot-path batch/import flow. Fixed the docblock wording though ("every enable path reachable from IPC", + the internal-enable note).
- **Post-crash `HOLOCHAIN_MANAGER` clear.** Real, but it needs the crash listener (in holochainManager) to signal index.ts to clear the global + notify the renderer — that's the same wiring as the renderer-surfacing TODO, so I'd rather do them together as the one follow-up than half of it here.
- **Polish (take-or-leave):** the `abortLaunch` helper, deleting the commented KanDo block, unifying the test harnesses, and deriving the drift floor — all reasonable, none blocking; leaving them per your "don't hold the PR for a fifth round." Same for the two altitude items (magic-string readiness, HolochainManager as the lifecycle chokepoint) and the earlier defers (`shutdown()`-awaits-exit, `processHandle` private + shared `stopSubprocesses()`, the bootstrap-guard question).

On `create-clone-cell` being first-party (so option (a) is one channel + one call, not new surface) — noted; if you'd rather close the clone window than document it, I'll take that as the follow-up.

Verified: `typecheck` + unit (145, incl. the new outage-path test) green; e2e #12 passes.

zippy (commented):
Verified 12cc3ef1 + f1807f5b against the head: maybeRebuild() is now the single rate limit and error boundary for all three authorize() branches — attempt-time gated, coalescing, never propagating (callers fail closed; the new outage test at appletZomeCallAuthorizer.test.ts:246 covers exactly the case round four flagged) — and invalidate() resetting lastAttemptAt to -Infinity preserves the immediate-pickup guarantee after install/uninstall. The leading await Promise.resolve() closes the sync-throw wedge, the ready branch re-checks settled and closes the admin websocket on both the timer-race and catch paths, the stdin error handler / 10-min inactivity threshold / launch-handler invalidate / LOCAL_SERVICES kill are all in, and the comment nits and the long-lived double blank line are gone.

Both pushbacks are accepted with the reasoning given. On enable: agreed — with the attempt-gated rebuild, a background refresh would reintroduce a sub-interval false-refuse window for a just-enabled cell, and the hard invalidate plus the lastAttemptAt reset is what guarantees the next call picks it up; that interaction is a better argument than the one I made against it. On the post-crash manager clear: pairing it with the renderer-surfacing TODO is the right scope.

Nothing further from me — this is ready to merge from a review standpoint. The follow-up list (polish items plus the two altitude threads) is well captured in the round-four thread.

PR #234, opened 2026-08-11
fix(e2e): disambiguate invite-dialog copy-link locator (fixes smoke #03)

zippy:
Smoke #03 (join-group) has been red on main-0.7 since the invite-code feature landed: the invite dialog now renders **two** `.copy-link-input` elements — the invite link and, below it, the invite code — so `getCurrentGroupInviteLink`'s locator matched two elements and failed Playwright strict mode (`strict mode violation: ... resolved to 2 elements`).

Fix: take `.first()` (the link). The helper's existing `expect.poll(readLink).toMatch(/invite/i)` already confirms it's the link, not the code.

This is independent of the Phase 1 PR (#233) — that PR inherits the same red because it branched off main-0.7; merging this un-reds the smoke suite for both.

### DIFF
--- a/tests/e2e/helpers/groups.ts
+++ b/tests/e2e/helpers/groups.ts
@@ -236,9 +236,12 @@ export async function getCurrentGroupInviteLink(page: Page): Promise<string> {
   // ("Node is not an <input>"). Read its `.value` JS property via evaluate.
   // Poll because the template binds `.value=...` reactively — first render
   // frame may have an empty value before invitationUrl resolves.
+  // The dialog now has two copy-link inputs — the invite link and, below it, the
+  // invite code. Take the first (the link); the /invite/ poll below confirms it.
   const readLink = () =>
     page
       .locator('invite-people-dialog sl-input.copy-link-input')
+      .first()
       .evaluate((el) => (el as HTMLElement & { value: string }).value);
   await expect.poll(readLink, { timeout: 10_000 }).toMatch(/invite/i);
   const link = await readLink();
--- a/tests/e2e/helpers/groups.ts
+++ b/tests/e2e/helpers/groups.ts
@@ -236,13 +236,17 @@ export async function getCurrentGroupInviteLink(page: Page): Promise<string> {
   // ("Node is not an <input>"). Read its `.value` JS property via evaluate.
   // Poll because the template binds `.value=...` reactively — first render
   // frame may have an empty value before invitationUrl resolves.
-  // The dialog now has two copy-link inputs — the invite link and, below it, the
+  // The dialog has two copy-link inputs — the invite link and, below it, the
   // invite code. Take the first (the link); the /invite/ poll below confirms it.
+  // A short evaluate timeout keeps a never-opening dialog (zero matches) failing
+  // fast with a locator error rather than hanging to the suite timeout.
   const readLink = () =>
     page
       .locator('invite-people-dialog sl-input.copy-link-input')
       .first()
-      .evaluate((el) => (el as HTMLElement & { value: string }).value);
+      .evaluate((el) => (el as HTMLElement & { value: string }).value, undefined, {
+        timeout: 5000,
+      });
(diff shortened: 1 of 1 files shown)


tests/e2e/helpers/groups.ts
```diff
@@ -236,9 +236,12 @@ export async function getCurrentGroupInviteLink(page: Page): Promise<string> {
   // ("Node is not an <input>"). Read its `.value` JS property via evaluate.
   // Poll because the template binds `.value=...` reactively — first render
   // frame may have an empty value before invitationUrl resolves.
+  // The dialog now has two copy-link inputs — the invite link and, below it, the
+  // invite code. Take the first (the link); the /invite/ poll below confirms it.
   const readLink = () =>
     page
       .locator('invite-people-dialog sl-input.copy-link-input')
+      .first()
```
zippy, line 246: Three hazards in the positional pick, one of which can silently return the wrong link:

1. **The selector is page-wide, not dialog-scoped.** `invite-people-dialog` is rendered by both `group-home` and `group-area-sidebar`, and `main-dashboard` keeps one hidden `group-container` per group in the DOM — so even the single-group case matches two dialogs (four inputs), and with ≥2 groups `.first()` resolves to whichever group's dialog comes first in DOM order, not necessarily the selected group. The `/invite/i` poll passes on *any* group's link, so a future multi-group caller gets a silent false green (agent 2 joins the wrong group; `toHaveCount(1)` still passes). At minimum, scope to the open dialog: `page.locator('invite-people-dialog:has(sl-dialog[open]) sl-input.copy-link-input').first()`.

2. **The `/invite/i` poll discriminates link-from-code only by accident** — the code format is `moss-<v>-<base64url>` and happens not to contain "invite". If the inputs are ever reordered (plausible: the code exists precisely because links get stripped in transit) the poll times out at 10s pointing at the helper, not the template; and since 546dadb7 the join dialog accepts codes too, so a caller without the poll would join successfully on the wrong artifact and nothing would notice.

3. **`readLink()` runs again after the poll**, so the returned value is not the value that satisfied it — a Lit re-render between the two reads can hand back `''` or a different match, surfacing 100+ lines later as "join dialog never closed". Capture the value inside the poll.

The right-altitude fix is in the dialog rather than the helper: both inputs share `class="moss-input copy-link-input"` and neither has a `.label` — unlike `join-group-dialog`'s `.label=${msg('Invite Link or Code')}`, which is exactly what lets `joinGroupByInviteLink` target semantically at line 207. Giving each input a label (`Invite link` / `Invite code`) fixes the screen-reader gap and lets this helper use `getByLabel` per the suite's own locator-priority rule (tests/e2e/README.md), making reorders and additions a no-op. Fine to land the `.first()` now and do the dialog labels as a follow-up.

zippy (commented):
As an unblocking one-liner this is fine to merge — it matches today's DOM and un-reds the smoke suite. The inline comment covers what the positional pick cements and where the real fix lives (the dialog, not the helper). Two smaller notes:

- `locator.evaluate` has no timeout (`actionTimeout` is unset in `playwright.config.ts`, so it waits unbounded) and `expect.poll` can't cancel an in-flight callback. If the dialog fails to open — zero matches — the first `readLink()` never resolves and the test hangs to the 240s suite timeout (900s in 09) instead of failing at 10s with a locator error. Passing `{ timeout: 5000 }` to `evaluate` restores a fast, pointed failure.
- The new comment's "The dialog now has two copy-link inputs" narrates the change (CLAUDE.md rule 6) — "has two" says the same thing without dating itself.

zippy (commented):
Re-checked 80416f04: the 5s evaluate timeout is passed correctly (third argument of evaluate, with arg=undefined) so a never-opening dialog now fails fast with a locator error instead of hanging to the suite timeout, and the comment no longer dates itself. Both notes from the previous review are resolved; the remaining suggestions (dialog-side labels, capturing the polled value, scoping to the open dialog) stand as follow-ups and none block merging. Merge order recommendation unchanged: this first, then #233.

PR #235, opened 2026-08-11
chore: bump profiles to v0.701.0 and group happ to 0.16.0-dev.3

zippy:
Pulls in the profiles fix (holochain-open-dev/profiles v0.701.0) where `create_profile`/`update_profile` no longer hit the network for the prefix-path ensure. HDI 0.8's `TypedPath` defaulted to `GetStrategy::Network`, so `path.ensure()` made a network `get_links` during profile creation — on a freshly joined group with the transport not yet registered on the relay this failed the whole zome call with:

```
Host("Connection attempted before home relay URL is known (src: None)")
```

Changes:
- `hc_zome_profiles_*` git deps → tag `v0.701.0` (coordinator crate 0.6.0)
- `@holochain-open-dev/profiles` → ^0.701.0 across workspaces
- group happ rebuilt: `moss.config.json` groupHapp 0.16.0-dev.3, sha256 `643563f7485bc8c208f170e718acf613d424f2d122a5fd7a3e671a4018725ce1`

Zome payloads for `create_profile`/`update_profile` are now `{ input: Profile, local?: bool }` (breaking, no 0.7 users yet). Moss call sites go through `ProfilesClient` and need no changes; the group DNA's profiles zomes re-export the crate wholesale.

Verified: zome wasm build, `hc app pack`, `yarn typecheck`, and the full group tryorama suite (11 tests).

Review follow-up (1507ee1a): three call sites hand-rolled `create_profile` payloads and would have broken against the new payload shape — dev-mode agent setup, group import/restore, and wdocker `join-group`. They now go through `ProfilesClient` (deep-imported from `dist/profiles-client.js`, added to the main-process bundle inline list), and the wdocker daemon's `get_agents_with_profile` calls moved behind the same client. Cargo.lock was regenerated with only the profiles tag change, dropping the unrelated itertools churn.

Note: the `@theweave/*` `0.7.0-dev.0 → dev.1` hunks in yarn.lock are stale-lockfile catch-up (those package.json versions were already dev.1 on the base branch), not part of this bump.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5815,7 +5815,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a56d757972c98b346a9b766e3f02746cde6dd1cd1d1d563472929fdd74bec4d"
 dependencies = [
  "anyhow",
- "itertools 0.13.0",
+ "itertools 0.14.0",
  "proc-macro2",
  "quote",
  "syn 2.0.101",
@@ -5828,7 +5828,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9120690fafc389a67ba3803df527d0ec9cbbc9cc45e4cc20b332996dfb672425"
 dependencies = [
  "anyhow",
- "itertools 0.13.0",
+ "itertools 0.14.0",
  "proc-macro2",
  "quote",
  "syn 2.0.101",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2395,8 +2395,8 @@ dependencies = [
 
 [[package]]
 name = "hc_zome_profiles_coordinator"
-version = "0.5.0"
-source = "git+https://github.com/holochain-open-dev/profiles?tag=v0.700.0#2d423491463b71bae3c97a506cced95d50d219ea"
+version = "0.6.0"
+source = "git+https://github.com/holochain-open-dev/profiles?tag=v0.701.0#7e11bd7e9380dce4583cc152115044ffa660fd19"
 dependencies = [
  "derive_more 0.99.20",
  "hc_zome_profiles_integrity",
@@ -2408,7 +2408,7 @@ dependencies = [
 [[package]]
 name = "hc_zome_profiles_integrity"
 version = "0.5.0"
-source = "git+https://github.com/holochain-open-dev/profiles?tag=v0.700.0#2d423491463b71bae3c97a506cced95d50d219ea"
+source = "git+https://github.com/holochain-open-dev/profiles?tag=v0.701.0#7e11bd7e9380dce4583cc152115044ffa660fd19"
 dependencies = [
  "derive_more 0.99.20",
  "hdi",
@@ -5815,7 +5815,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a56d757972c98b346a9b766e3f02746cde6dd1cd1d1d563472929fdd74bec4d"
 dependencies = [
  "anyhow",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "proc-macro2",
  "quote",
  "syn 2.0.101",
@@ -5828,7 +5828,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9120690fafc389a67ba3803df527d0ec9cbbc9cc45e4cc20b332996dfb672425"
 dependencies = [
  "anyhow",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "proc-macro2",
  "quote",
  "syn 2.0.101",
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@theweave/cli",
-  "version": "0.16.0-dev.3",
+  "version": "0.16.0-dev.4",
   "description": "CLI to run Tools for the Weave in development mode",
   "license": "CAL-1.0",
   "repository": {
@@ -24,7 +24,7 @@
     "@electron-toolkit/preload": "^2.0.0",
     "@electron-toolkit/utils": "^2.0.0",
     "electron-updater": "6.6.2",
-    "@theweave/api": "^0.7.0-dev.1",
+    "@theweave/api": "^0.7.0-dev.2",
     "@theweave/moss-types": "0.7.0-dev.1",
     "@theweave/utils": "0.7.0-dev.1",
     "@msgpack/msgpack": "^2.8.0",
--- a/cli/package.json
+++ b/cli/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "@holochain/client": "^0.21.0",
     "@holochain-open-dev/elements": "^0.700.0",
-    "@holochain-open-dev/profiles": "^0.700.0",
+    "@holochain-open-dev/profiles": "^0.701.0",
     "@holochain-open-dev/stores": "^0.700.0",
     "@holochain-open-dev/utils": "^0.700.0",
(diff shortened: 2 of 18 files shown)


zippy:
Review of the profiles 0.700.0 → 0.701.0 bump. The bump itself is correctly propagated — all 7 package.jsons and both zome Cargo.tomls updated, single profiles resolution in yarn.lock, and the published `group-happ-v0.16.0-dev.3` asset's sha256 matches `moss.config.json` byte-for-byte (verified by download; the wasm inside is built from tag `v0.701.0`, commit `7e11bd7e`). The integrity zome is unchanged between tags, so the DNA hash and network compatibility are unaffected.

The problem: v0.701.0 changed `create_profile`/`update_profile` to take `ZomeFnInput<Profile>` (`{ input, local }`) instead of a bare `Profile`, and three call sites bypass `ProfilesClient` with hand-rolled `callZome` payloads — so the bump breaks them. The root defect is the bypass, not the payloads; the fix is to route these through `ProfilesClient` so the wire format lives in one place.

### Findings

**1. `src/main/cli/devSetup.ts:497` — dev-mode profile creation breaks**

Raw `create_profile` call sends bare `{ nickname, fields }`. Against the dev.3 coordinator this fails deserialization ("missing field `input`"), the `await` is uncaught, and every `yarn applet-dev*` launch aborts during agent setup.

`ProfilesClient` works here as-is: it's a plain `ZomeClient` wrapper with no lit/DOM dependencies (deep-importable as `@holochain-open-dev/profiles/dist/profiles-client.js`), its constructor takes the `AppWebsocket` the code already holds (`groupWebsocket`), and `src/main` already depends on the package.

**2. `src/main/index.ts:2218` — group import silently stops setting profiles**

Same hand-rolled payload in the import/restore flow, and the surrounding `catch` only `console.warn`s — so the import reports success while the user ends up profileless in every imported group. Invisible in CI and to the user. Same fix: `ProfilesClient` (the file already imports the `Profile` type from the package).

**3. `wdocker/src/commands/conductor/join-group.ts:53` — wdocker join-group breaks**

Same bare payload, uncaught; wdocker downloads the happ at `MOSS_CONFIG.groupHapp.version` (dev.3), so `join-group` installs the group then exits with an error, leaving no `wdockerNode` profile for the daemon. wdocker doesn't depend on `@holochain-open-dev/profiles` — either add it (its `@holochain/client ^0.21.0` peer range is already satisfied) or house a profiles wrapper in `@theweave/group-client` next to the zome clients wdocker already uses. Note `wdocker/src/daemon/daemon.ts` also hand-rolls `get_agents_with_profile` — it happens to use the correct `{ input, local }` shape today, but it's the same drift risk and should move behind the same client.

**4. `Cargo.lock` (~5818) — unrelated itertools churn**

The regeneration flipped itertools 0.14.0 → 0.13.0 under prost-build/prost-derive. Both versions remain in the graph, so it dedupes nothing; no build impact, but it's diff noise and a latent merge conflict, and the next full re-resolution can flip it back. Regenerating the lock with only the profiles tag change drops these hunks.

### Notes

- The `@theweave/*` `0.7.0-dev.0 → dev.1` hunks in yarn.lock are stale-lockfile catch-up (those package.json versions were already `dev.1` on the base branch), not part of this bump — worth a line in the PR description so reviewers don't hunt for a dependency edge that doesn't exist.
- Groups installed under earlier dev builds keep the old 0.5.0 coordinator and would break with the new client — noted as accepted-by-design for the dev series, no action needed.

zippy:
All four addressed in 1507ee1a:

1. **devSetup.ts** — now uses `ProfilesClient.createProfile` (deep import of `dist/profiles-client.js`; `@holochain-open-dev/profiles` added to the main-process bundle inline list so the runtime import bundles instead of being externalized as ESM into the CJS main).
2. **index.ts import flow** — same, via the existing `appWs`.
3. **wdocker join-group** — added `@holochain-open-dev/profiles` as a direct dep (resolves to the existing 0.701.0 lockfile entry, no yarn.lock change) and routed through `ProfilesClient`; the daemon's three `get_agents_with_profile` calls moved behind `getAgentsWithProfile` as well.
4. **Cargo.lock** — regenerated from the base lock; the itertools flip reproduces deterministically with `cargo update -p`, so the two lines were restored to the base resolution by hand and verified stable under `cargo check` (both itertools versions were already in the graph, so the pre-existing resolution remains valid).

The yarn.lock stale-catch-up note is added to the PR description. Wasm and DNA are unaffected (no zome changes), so the published `group-happ-v0.16.0-dev.3` asset and sha stand. Verified: `yarn typecheck`, wdocker `tsc` build, `electron-vite build` (confirmed `ProfilesClient` inlined in `out/main`, no externalized require), prettier clean.

zippy:
Verified all four fixes in 1507ee1a:

1. **devSetup.ts / index.ts / join-group.ts** — all three sites now go through `ProfilesClient`; confirmed against the v0.701.0 client source that `createProfile(profile)` wraps into `{ input, local }` with `local` defaulting to `true`. The deep import of `dist/profiles-client.js` is valid under the package's exports map, and the main-process inline-list addition in `electron.vite.config.ts` is consistent with how `@holochain-open-dev/utils` is already handled.
2. **wdocker** — the "no yarn.lock change" claim holds (wdocker is a workspace and its new `^0.701.0` range matches the existing lockfile key). The daemon's `getAgentsWithProfile(true/false)` calls send payloads byte-identical to the ones they replaced, including the local-first/network-refresh ordering.
3. **Cargo.lock** — both prost-derive entries restored to `itertools 0.14.0`, matching the base resolution; the lock delta is now exactly the profiles tag change.

### Remaining follow-up: the published @theweave packages

This PR bumps the profiles dependency in three **npm-published** packages without bumping their own versions, so the dep update cannot reach npm:

- **`@theweave/api`** and **`@theweave/elements`**: repo versions stay at `0.7.0-dev.1`, which is already published (`latest`) declaring `@holochain-open-dev/profiles: ^0.700.0`. Since a caret range on a 0.x version pins the minor, `^0.700.0` *excludes* 0.701.0 — downstream installs keep resolving the old client. Both need a version bump (`0.7.0-dev.2`) and publish. Runtime exposure is limited (api's profiles import is type-only, and applet UIs receive the client *instance* from the applet-iframe, which this PR does update), but the published metadata should not keep pointing tool developers at the pre-fix client.
- **`@theweave/cli`**: the repo says `0.16.0-dev.3`, but npm's `0.16.0-dev.3` is a different artifact — built pre-PR, bundling the old raw `create_profile` code and a `moss.config.json` pinned to group happ `0.16.0-dev.2`. That published CLI is internally consistent (old client + old happ), so nothing is broken on npm today, but the next publish must be `0.16.0-dev.4`, and until it ships, tool developers cannot develop against the dev.3 group happ / the local-first profiles behavior this whole bump delivers.

Suggest either folding the three version bumps into this PR or filing a tracked follow-up so the publishes don't get lost.

zippy:
Version bumps folded into this PR (17aa4e8e): `@theweave/api` and `@theweave/elements` → `0.7.0-dev.2`, `@theweave/cli` → `0.16.0-dev.4`, with elements and cli declaring `@theweave/api ^0.7.0-dev.2` so the published metadata points at the fixed stack. yarn.lock is unchanged (workspace-internal bumps, ranges already satisfied).

Full `yarn build` passes including `build:cli`; the assembled `cli/dist` bundles the ProfilesClient-routed main process and a `moss.config.json` pinned to group happ `0.16.0-dev.3` / sha `643563f7…25ce1`, so the published CLI will be internally consistent with the new happ.

Publishes are pending (OTP-gated): `npm publish` from `libs/api`, then `libs/elements`, then `cli` — dev-line releases go to the `latest` dist-tag, so no `--tag` needed.

PR #248, opened 2026-09-23
Audio source capture: Moss host side and @theweave/api captureAudioSources

zippy:
## Audio source capture: Moss host side and `@theweave/api` helper

A Tool can ask Moss for audio that is playing on the user's machine. Moss shows its own picker (all system output, or individual applications, playing ones first), captures the chosen sources through `@lightningrodlabs/flexaudio` with Moss's own process tree excluded, and hands the Tool a `MessagePort` of 20 ms mono 48 kHz Int16 frames. `@theweave/api` turns that port into a live `MediaStreamTrack` through an AudioWorklet. Design: Presence repo, `docs/superpowers/specs/2026-09-22-audio-source-capture-design.md` (Sections 1 to 3 landed; Section 4, the Presence side, follows).

### What is in this PR

- **Main process**: `audioCapture.ts` (lazy addon load + capability probe), `audioMixer.ts` (pure frame mixer), `audioSourceGrants.ts` (grant engine over injected bindings: picker rows, one capture stream per source, a wall-time-gated 20 ms pump, one idempotent `endGrant`), `audioSourcePicker.ts`, `audioSourcesIpc.ts` (six IPC channels, paired by the existing drift test), one construction block in `index.ts`.
- **Preloads**: the bridge methods and the `MessagePort` relay into page script.
- **Renderer**: the `request-audio-sources` Weave message in both the main window and WAL windows (reply carries the port as a transferable), a per-grant chip, Settings → Services → Audio Sources (allow switch, capability readout, active grants), eight translations.
- **`@theweave/api` 0.7.0-dev.4**: optional `captureAudioSources()` on `WeaveServices`/`WeaveClient` (feature-detected), `createAudioSourceCapture`, `PcmRing`, `CaptureSession`; the applet iframe's port-aware reply path; README section; regenerated API reference.
- **Dependency**: `@lightningrodlabs/flexaudio` pinned exactly at `0.3.0-lrl.1`; `NSAudioCaptureUsageDescription` for macOS.

### Verified

- `yarn test:unit` 35 files / 375 tests, `yarn typecheck` green at HEAD.
- Live, against a real 0.7 conductor over CDP: picker → grant → frames (peak 0.80 with a tone) → chip Stop → `ended`; cancel → `null`; switch off → `null`; iframe and main-window reload both release the grant; the Tool-side helper round trip from the example applet.
- Measured: the host pump emits 50.1 frames/s and the Tool ring shows zero overflow over 20 s (an earlier build over-emitted ~5%; the keepalive hypothesis was measured and falsified before the pump was fixed).

### Not in this PR / owner steps

- `npm publish` of `@theweave/api@0.7.0-dev.4` (`cd libs/api && npm publish --tag latest`; `latest` currently points at `0.7.0-dev.3`).
- The WAL-window path was not observed live: the example applet's committed `.happ` predates its zome source.
- Known limits recorded in the spec: `canExcludeSelf` is a request-time snapshot (macOS resolves pids once at start); a host stall over 5 s ends the capture on both ends by design.


zippy (commented):
Review of `138c8a3b` against `main-0.7`. Two confirmed capture leaks, three plausible teardown gaps, and four cleanups. Details are in the inline comments.

**Fix before merge**
- `audioSourceGrants.ts:285`: stream ids come from the map size, so a row that fails synchronously while opening hands its id to the next row. A late `deviceLost` for the dead stream then closes the wrong entry and ends the grant while the live native stream keeps capturing.
- `audio-source-capture.ts:96`: the worklet source instantiates the ring class by its runtime `.name`. A Tool bundler that renames the class binding gets a silent track with no error.

**Plausible teardown gaps** (same shape: a main-process grant outlives every handle that can stop it)
- `grants-client.ts:66`: grants pending a port, or keyed `'unregistered'`, are invisible to `endForIframe`.
- `applet-iframe/src/index.ts:334`: a reply without a transferred port throws on the Tool side but the grant keeps running.
- `walwindow.ts:310`: an awaited `endForIframe` that rejects aborts the `unregister-iframe` forward.

**Cleanups**
- `src/main/index.ts:1411`: the native addon loads eagerly at startup although the engine takes a thunk.
- `audioSourceGrants.ts:171`: the process list is enumerated twice per picker open.
- `audioSourceGrants.ts:296` and the `pump` docblock: comments narrate prior behavior (CLAUDE.md rule 6).
- `api.ts:490` and `audioSourcePicker.ts:34`: duplicated `hostCapture` wiring and a duplicated one-picker guard.

Generated docs, XLIFF files and `yarn.lock` were not reviewed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

src/main/audioSourceGrants.ts
```diff
@@ -0,0 +1,398 @@
...
+    await Promise.all(ids.map((id) => this.endGrant(id, reason)));
+  }
+
+  private openStream(
+    grant: Grant,
+    backend: AudioCaptureBackend,
+    kind: 'system' | 'app',
+    options: OpenOptions,
+  ): boolean {
+    const id = `${kind}-${grant.streams.size}`;
```
zippy: Stream ids are derived from `grant.streams.size`. A stream that fails synchronously inside `backend.openStream` is deleted by `closeStream`, and the next row is assigned the same id.

Scenario: two app rows, no system row. Row 1 opens as `app-0` and fires a synchronous `error`, so `closeStream` deletes `app-0`. Row 2 gets `app-0` again and succeeds. The addon later emits `deviceLost` for the closed row-1 stream. `closeStream(grant, stream1)` deletes `app-0`, which is now row 2's entry, stops row 1's already-stopped handle, and since the map is empty ends the grant as `app-quit`. Row 2's native stream is never stopped and keeps capturing.

Use a per-grant monotonic counter for ids.

libs/api/src/audio-source-capture.ts
```diff
@@ -0,0 +1,332 @@
...
+ * node port, each render quantum pulls from the ring, `{type:'stats'}` asks
+ * for the counters and `{type:'close'}` retires the processor.
+ * @public
+ */
+export const WORKLET_SOURCE = `
+${PcmRing.toString()}
+class MossAudioSourceProcessor extends AudioWorkletProcessor {
+  constructor() {
+    super();
+    this.ring = new ${PcmRing.name}(${RING_CAPACITY_SAMPLES});
```
zippy: `WORKLET_SOURCE` splices `PcmRing.toString()` but instantiates it as `new ${PcmRing.name}(...)`. That only works when the bundler keeps the class binding name equal to its `.name`.

With esbuild `keepNames`, or any transform that emits `var X = class { ... }`, `PcmRing.name` is `"PcmRing"` but `toString()` yields `class X {` or `class {`. The worklet module then references a binding that does not exist. `addModule` and `new AudioWorkletNode` succeed, the processor constructor throws (`processorerror`), `process()` never runs, and `createAudioSourceCapture` resolves with a silent track and no error.

Bind the class expression to a fixed local, for example `const Ring = ${PcmRing.toString()};` and `new Ring(...)`, so the name is irrelevant.

src/renderer/src/audio-sources/grants-client.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+    this.b.armPortDeadline(requestId);
+    let port: MessagePort;
+    try {
+      port = await portPromise;
+    } catch (e) {
+      await this.b.stopAudioSources(result.grantId, 'iframe-unloaded');
+      throw e;
+    }
+    this.byIframe.set(req.iframeKey, [...(this.byIframe.get(req.iframeKey) ?? []), result.grantId]);
```
zippy: A grant is only recorded in `byIframe` after the port arrives, and iframes that `findIframeIdBySource` cannot match go under the `'unregistered'` key that nothing ever ends.

If the iframe sends `unregister-iframe` while `request()` is between the resolved invoke and `await portPromise`, or was keyed `'unregistered'`, `endForIframe(id)` finds nothing and the main-process capture runs until the orphaned MessagePort is garbage-collected.

Record the pending grantId as soon as the invoke resolves, and stop grants keyed `'unregistered'` instead of dropping them.

iframes/applet-iframe/src/index.ts
```diff
@@ -322,6 +325,19 @@ const weaveApi: WeaveServices = {
       type: 'user-select-screen',
     }),
 
+  captureAudioSources: async (
+    opts?: CaptureAudioSourcesOptions,
+  ): Promise<AudioSourceCapture | null> => {
+    const { result, ports } = await postMessageWithPorts({ type: 'request-audio-sources' });
+    if (!result) return null;
+    const port = ports[0];
+    if (!port) throw new Error('The host granted audio sources but transferred no port.');
```
zippy: If the host answers with a result but `m.ports` is empty, `captureAudioSources` rejects, but the main-process grant that was just created keeps capturing. The reply carries no grantId, so nothing on the Tool side can stop it.

Either include `grantId` in the reply so the iframe can request a stop, or have the host-side handler treat a missing port as a failed delivery and end the grant.

src/renderer/src/walwindow.ts
```diff
@@ -282,6 +307,7 @@ export class WalWindow extends LitElement {
                   request.request.id,
                 );
               }
+              await audioSourceGrantsClient.endForIframe(request.request.id);
```
zippy: `endForIframe` is awaited before the `unregister-iframe` forward. If the `stop-audio-sources` invoke rejects (IPC error, or main is stuck in a native `stop()`), the handler throws before `appletMessageToParent(...)` runs and the main window keeps a stale iframe entry.

The same shape is in `applet-host.ts` around lines 470 and 477, where the throw replaces the reply. Fire-and-forget or try/catch the grant teardown so unregister always completes.

src/main/index.ts
```diff
@@ -1401,6 +1408,42 @@ if (!RUNNING_WITH_COMMAND) {
         };
       });
     });
+    const audioBackend = loadAudioCapture();
```
zippy: `loadAudioCapture()` runs synchronously at startup and dlopens the native addon (which links libpipewire) before the app is usable. The engine already takes `backend` as a thunk.

Make the thunk load on first call, for example `let backend; () => (backend ??= loadAudioCapture())`, so a launch that never asks for audio does no audio work.

src/main/audioSourceGrants.ts
```diff
@@ -0,0 +1,398 @@
...
+    }));
+  }
+
+  async request(req: AudioSourceRequest): Promise<AudioSourceRequestResult | null> {
+    const backend = this.b.backend();
+    const caps = await probeAudioCapabilities(backend, this.b.platform);
+    if (!backend || !caps.supported) return null;
+
+    const excludePids = this.b.excludePids();
+    const processes = caps.perApp ? await backend.processes().catch(() => [] as JsProcessInfo[]) : [];
```
zippy: Each request enumerates the process list twice: `probeAudioCapabilities` calls `backend.processes()` and discards the result, then `request` calls it again for the rows. On PipeWire each call is a synchronous graph walk.

Have the probe return the process list, or accept a pre-fetched one, and build the rows from it.

src/main/audioSourceGrants.ts
```diff
@@ -0,0 +1,398 @@
...
+    const onChunk = (chunk: JsAudioChunk) => {
+      if (chunk.frames === 0) return;
+      stream.queue.push(chunk.data);
+    };
+    const onEvent = (ev: JsStreamEvent) => this.onStreamEvent(grant, stream, ev);
+    // Registered before the backend call: a fatal event the addon fires
+    // synchronously from inside openStream (permissionDenied/error at open
+    // time) reaches onStreamEvent -> closeStream while the addon call is
+    // still on the stack, and closeStream can only tear the stream down if
+    // it can find it in grant.streams. Registering after the call returned
```
zippy: This comment and the `pump` docblock (lines 362-364) describe the previous behavior and the measurement that motivated the change. CLAUDE.md rule 6 asks for comments that state intent only. Keep the invariant (the entry must exist before the addon can fire a synchronous event) and move the history to the commit message.

libs/api/src/api.ts
```diff
@@ -478,7 +486,10 @@ export class WeaveClient implements WeaveServices {
         window.__WEAVE_APPLET_SERVICES__ = appletServices;
       }
       window.dispatchEvent(new CustomEvent('weave-client-connected'));
-      return new WeaveClient();
+      const client = new WeaveClient();
+      const hostCapture = window.__WEAVE_API__?.captureAudioSources;
```
zippy: The `hostCapture` detection and wiring is duplicated in both branches of `connect()` (here and at 506-507). Move it into the `WeaveClient` constructor or a `fromHost()` helper so the next optional host method is wired once.

src/main/audioSourcePicker.ts
```diff
@@ -0,0 +1,60 @@
...
+  picker.window.close();
+}
+
+/**
+ * Opens the audio-source picker and resolves the chosen row ids, or null when
+ * the user cancels or closes the window. One picker at a time, like the
+ * screen/window picker.
+ */
+export function openAudioSourcePicker(rows: AudioSourceRow[]): Promise<string[] | null> {
+  if (PICKER) return Promise.reject(new Error('Only one audio source picker may be open at a time.'));
```
zippy: This guard duplicates `pickerOpen` in `AudioSourceGrants.request` (`audioSourceGrants.ts:174`) with the same error text. The engine's flag is the one the tests exercise, and this check is unreachable through the engine. Keep the check in one layer.

iframes/applet-iframe/src/index.ts
zippy: Not changing the reply shape. With a Moss host the port is attached by the same reply that carries the non-null result: `TransferableReply` in `applet-host.ts` and `walwindow.ts` posts `{type:'success', result}` with `[grant.port]` as the transfer list, so a result without a port cannot be produced by this host; the throw guards a foreign host that answers the message without honouring the transfer. Putting `grantId` in the reply would let a Tool name grants it does not own (`stop-audio-sources` takes a bare grantId), which the reply shape deliberately withholds. A comment at this site now states the branch is unreachable with a Moss host.

src/main/audioSourceGrants.ts
zippy: Fixed in 89819869. Stream ids come from a per-grant monotonic counter (`nextStreamSeq`), so a row that fails synchronously while opening never hands its id to the next row. Pinned by a test that reproduces the scenario (row 1 fires a synchronous `error` on open, row 2 succeeds, a late `deviceLost` for the dead row-1 stream arrives): the grant stays live with row 2's stream, and row 2's handle is not stopped; a negative control checks a late `deviceLost` on a live stream still ends it.

libs/api/src/audio-source-capture.ts
zippy: Fixed in 382b0f48. The module now binds the spliced class to a fixed local (`const Ring = <class text>;` and `new Ring(...)`), built by an internal `buildWorkletSource(ringClassSource)`, so the binding name is irrelevant. Tests evaluate the module with an anonymous class and with the class renamed to `Q`; both still pull samples through `process()`, and the module text is asserted not to reference the class by name.

src/renderer/src/audio-sources/grants-client.ts
zippy: Fixed in f8fb8a22. The grantId is recorded under the iframe key as soon as the invoke resolves, before the port is awaited; an `endForIframe` that races in during that window stops the grant, and `request` then closes the late port and resolves `null` instead of handing back a stopped grant. The `'unregistered'` key is gone: a request from an iframe `findIframeIdBySource` cannot match now throws before main is invoked, so no grant is created (every applet iframe registers via `get-iframe-config` before it can send anything else).

src/renderer/src/walwindow.ts
zippy: Fixed in f8fb8a22. All three sites (`applet-host.ts` both arms, `walwindow.ts`) now fire-and-forget the grant release with the failure logged, so `unregister-iframe` always forwards/replies.

src/main/index.ts
zippy: Fixed in 89819869. The engine's `backend` thunk now loads the addon on first call, one attempt (a host without PipeWire warns once and stays `supported:false`); the capabilities IPC uses the same thunk. A launch that never asks for audio does no audio work.

src/main/audioSourceGrants.ts
zippy: Fixed in 89819869. `probeAudioSupport` returns `{ capabilities, processes }` from one `devices()`/`processes()` pass; `request()` builds the picker rows from that list and `probeAudioCapabilities` (the settings readout) derives from the same probe. A test asserts exactly one `processes()` call per request.

src/main/audioSourceGrants.ts
zippy: Fixed in 89819869. The register-before-open comment now states only the invariant; the measurement sentences are gone from the `pump()` docblock (they live in the commit messages).

libs/api/src/api.ts
zippy: Fixed in 382b0f48. Host-feature detection is wired once (a single helper called from both `connect()` branches); the optional-property semantics are unchanged.

src/main/audioSourcePicker.ts
zippy: Fixed in 89819869. The picker module's guard is removed; the engine's `pickerOpen` in `AudioSourceGrants.request` is the one authority (and the one the tests exercise), and `openAudioSourcePicker` documents that callers serialise.

zippy:
Addressed the review in three commits: 89819869 (main), 382b0f48 (api), f8fb8a22 (renderer); one item pushed back in-thread. Gate at f8fb8a22: 35 files / 388 tests, typecheck green. Each thread has its own reply.

iframes/applet-iframe/src/index.ts
zippy: Accepted. Both host handlers build the reply as a `TransferableReply` with the port in the transfer list, so a non-null result without a port cannot come from Moss, and the reply shape withholding `grantId` is the right call given `stop-audio-sources` takes a bare id. The comment at the throw site is enough.

zippy (commented):
Re-review of `f8fb8a22` (89819869, 382b0f48, f8fb8a22 on top of `138c8a3b`). All nine fixed items are confirmed in the code, and the one pushback is accepted (reply in thread).

Verified on this head with a fresh install and `yarn build:libs`:
- `vitest run`: 35 files, 388 tests, all passing.
- `typecheck:node` and `typecheck:web`: clean.

Notes per item:
- Stream ids: `nextStreamSeq` on the grant; the new test opens two app rows, fails the first synchronously, fires a late `deviceLost` for it, and asserts the second stream survives and keeps pumping. The negative control still ends a live stream. Matches the reported scenario exactly.
- Worklet ring binding: `buildWorkletSource` binds the spliced class to `const Ring`; tests evaluate the module with an anonymous and a renamed class and pull samples through `process()`.
- Pending grants: `record` runs as soon as the invoke resolves, `holds` decides whether the late port is handed over or closed, and the `'unregistered'` key is gone in both windows. The refusal for an unmatched iframe is a throw before main is invoked, which posts an error back to the iframe; fine.
- Unregister forward: `releaseGrantsFor` is fire-and-forget with the failure logged at all three sites.
- Lazy addon, single enumeration, comment cleanup, `fromHost()`, and the single picker guard: all as described.

No further changes requested.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

